PR #1, opened 2023-06-08
Project setup and common modules

MancunianSam:


common.tf
```diff
@@ -0,0 +1,71 @@
...
+  source                       = "git::https://github.com/nationalarchives/da-terraform-modules//vpc"
+  vpc_name                     = "${local.environment}-vpc"
+  az_count                     = local.az_count
+  elastic_ip_allocation_ids    = [aws_eip.eip.allocation_id]
+  nat_instance_security_groups = [module.nat_instance_security_group.security_group_id]
+}
+
+resource "aws_eip" "eip" {
+  domain = "vpc"
+}
```
kurtismash, line 38: DOes 1 EIP allow us to be HA?

kurtismash (commented):
A couple of comments.

common.tf
MancunianSam: Good point, I've put a count in with the az_count.

templates/secrets_manager/secrets_manager_lambda_policy.json.tpl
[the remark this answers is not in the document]
MancunianSam: I thought you needed a secrets manager policy to allow services to access it but I don't think you do so I've deleted this.

kurtismash (approved):

PR #2, opened 2023-06-09
Download metadata lambda

MancunianSam:
Add the download metadata and files lambda and associated queues, topics etc.
Also added a test and a deploy workflow for GitHub actions. The tests won't pass until https://github.com/nationalarchives/dr2-github-actions/pull/1 is merged.


download-metadata-and-files.tf
```diff
@@ -0,0 +1,56 @@
+locals {
+  disaster_recovery_bucket_name = "${local.environment}-disaster-recovery"
+  lambda_name                   = "${local.environment}-download-files-metadata"
+  queue_name                    = "${local.environment}-download-files-metadata"
```
kurtismash, line 4: Should these contain the word `and`?

download-metadata-and-files.tf
```diff
@@ -0,0 +1,56 @@
...
+  source                   = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
+  queue_name               = local.queue_name
+  sqs_policy               = templatefile("./templates/sqs/sqs_access_policy.json.tpl", { role_arn = module.download_metadata_and_files_lambda.lambda_role_arn, account_id = var.dp_account_number, queue_name = local.queue_name })
+  redrive_maximum_receives = 3
+  tags = {
+    CreatedBy = "dp-terraform-environments"
+  }
+}
+
+module "disaster_recovery_bucket" {
```
kurtismash, line 48: Maybe we should break this out? This bucket will also be needed by our DR workflow on-prem - of which they'll be some more resources based on it.

templates/deploy_lambda_policy.json.tpl
```diff
@@ -0,0 +1,18 @@
...
+      "Action": [
+        "lambda:GetFunctionConfiguration",
+        "lambda:PublishVersion",
+        "lambda:UpdateFunctionCode",
+        "s3:GetObject",
+        "s3:PutObject"
+      ],
+      "Effect": "Allow",
+      "Resource": [
+        "${download_metadata_lambda_arn}"
```
kurtismash, line 13: No0b question - does this need an S3 resource as well?

templates/deploy_lambda_policy.json.tpl
kurtismash, line 1: Is this file a duplicate?

download-metadata-and-files.tf
```diff
@@ -0,0 +1,56 @@
...
+  tags = {
+    Name      = local.lambda_name
+    CreatedBy = "dp-terraform-environments"
+  }
+}
+
+module "download_files_sqs" {
+  source                   = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
+  queue_name               = local.queue_name
+  sqs_policy               = templatefile("./templates/sqs/sqs_access_policy.json.tpl", { role_arn = module.download_metadata_and_files_lambda.lambda_role_arn, account_id = var.dp_account_number, queue_name = local.queue_name })
```
kurtismash, line 41: Why does this lambda need to send to SQS?

kurtismash (commented):
Looks good, a few comments. Also is it worth renaming the `.tf` files for lambdas to make it clear these are lambda functions. An alternative could be grouping lambdas in files based on their workflow? I.e. one file for the event generators, one for the DR workflow, one for the closure workflow, etc.

download-metadata-and-files.tf
MancunianSam: I suppose so, I've added it in.

download-metadata-and-files.tf
MancunianSam: Yeah, fair point. I've put it into a different file.

templates/deploy_lambda_policy.json.tpl
MancunianSam: Yep, it definitely does. I've added one in.

templates/deploy_lambda_policy.json.tpl
MancunianSam: It is, I've deleted it.

download-metadata-and-files.tf
MancunianSam: It doesn't need to send to SQS. This is supposed to be the queue that triggers the lambda. I've got it the wrong way round. I've changed it now so it is the queue that triggers the lambda.

templates/sqs/sqs_access_policy.json.tpl
kurtismash, line 1: Is this policy not too permissive? Applying the principal of least privilege here should only permit the SNS topic to `sqs:SendMessage` to the queue.

templates/deploy_lambda_policy.json.tpl
kurtismash: As we're doing object functions I think we need the `/*` on the end of the bucket resource.

kurtismash (approved):

MancunianSam:
I've also set up terraform to use an external ID from the da-terraform-configurations repo.

templates/sqs/sqs_access_policy.json.tpl
kurtismash: Is it worth limiting the principal as well?

MancunianSam: I've put a TODO in to change it when the SNS topic exists.

kurtismash (approved):

PR #4, opened 2023-06-22
Add slack notifications lambda module

techncl:


common.tf
```diff
@@ -14,6 +14,11 @@ resource "aws_secretsmanager_secret" "preservica_secret" {
   name = "${local.environment}-preservica-api-login-details-${random_string.preservica_user.result}"
 }
 
+resource "aws_secretsmanager_secret" "slack_webhook_url" {
```
MancunianSam, line 17: Sorry to be a pain but can you put this in SSM parameter store instead of secrets manager. 

Parameter store is a lot cheaper and just as secure if you use SecureString. We've already got quite a bit in parameter store as well. 

Secrets manager makes sense for preservica because we can change the credentials programatically so the rotation features are worth it but the slack webhook will be updated manually, if at all. 

You can add it as a resource, there is a module but that's more for creating lots of parameters at once.

```
resource "aws_ssm_parameter" "slack_webhook" {
  name = "/${local.environment}/slack/webhook"
  type = "SecureString"
}
```

You'll have to change your lambda code to get it from SSM instead

templates/iam_policy/slack_notifications_policy.json.tpl
```diff
@@ -0,0 +1,24 @@
+{
+  "Statement": [
+    {
+      "Action": "secretsmanager:GetSecretValue",
```
MancunianSam, line 4: You'll need to update this to use ssm:GetParameter and ${parameter_arn}

slack_notifications_lambda.tf
```diff
@@ -0,0 +1,29 @@
...
+}
+
+module "slack_notifications_lambda" {
+  source        = "git::https://github.com/nationalarchives/da-terraform-modules//lambda"
+  function_name = local.notifications_lambda_name
+  handler       = "lambda_function.lambda_handler"
+
+  policies = {
+    "${local.notifications_lambda_name}-policy" = templatefile("./templates/iam_policy/slack_notifications_policy.json.tpl", {
+      secrets_manager_secret_arn   = "arn:aws:secretsmanager:eu-west-2:${var.dp_account_number}:secret:slack_webhook_url-xrHr13",
```
MancunianSam, line 12: This will need to change to use the ssm parameter arn. You shouldn't need to construct the arn, you should be able to do `aws_ssm_parameter.slack_webhook.arn`

MancunianSam (commented):
Looks good mate. I know switching to SSM is a faff but I'd like to restrict using secrets manager to things where the rotation features are useful.

common.tf
techncl: Sure. Sorry, I misread; thought the card said "Secrets Manager"🤦

MancunianSam:
Looks good. The terraform modules for the VPC have been updated so you need to add an environment parameter to get the tests to pass like this https://github.com/nationalarchives/dp-terraform-environments/blob/9ba9ea5dd0d07f56b9b99f831ab9bf6710d3fefd/common.tf#L31

MancunianSam (approved):
Looks good, thanks.

PR #9, opened 2023-07-04
DR2-1093 Preservica Config Lambda

MancunianSam:
Add Preservica config lambda.
    
I've used this PR to update the existing S3 buckets and SQS queues to use KMS.
There are two keys. dr2_kms_key only has the download_metadata_files lambda role allowed to encrypt and decrypt.
dr2_developer allows the developer admin roles to encrypt and decrypt.
This will be used for config buckets and queues like preservica config and the deploy bucket.

The tests won't pass on this until https://github.com/nationalarchives/da-terraform-modules/pull/9 is merged but this is ready for review.

### DIFF
--- a/deploy_preservica_config.tf
+++ b/deploy_preservica_config.tf
@@ -30,6 +30,9 @@ module "preservica_config_sns" {
   }
   topic_name  = "${local.environment}-preservica-config"
   kms_key_arn = module.dr2_developer_key.kms_key_arn
+  sqs_subscriptions = {
+    preservica_config_queue = module.preservica_config_queue.sqs_arn
+  }
 }
 
 module "preservica_config_queue" {
--- a/deploy_preservica_config.tf
+++ b/deploy_preservica_config.tf
@@ -38,8 +38,10 @@ module "preservica_config_sns" {
 module "preservica_config_queue" {
   source     = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
   queue_name = "${local.environment}-preservica-config"
-  sqs_policy = templatefile("${path.module}/templates/sqs/sqs_access_policy.json.tpl", {
+  sqs_policy = templatefile("${path.module}/templates/sqs/sns_send_message_policy.json.tpl", {
     account_id = data.aws_caller_identity.current.account_id, queue_name = "${local.environment}-preservica-config"
+    queue_name = "${local.environment}-preservica-config"
+    topic_arn  = module.preservica_config_sns.sns_arn
   })
   kms_key_id = module.dr2_developer_key.kms_key_arn
 }
--- a/deploy_preservica_config.tf
+++ b/deploy_preservica_config.tf
@@ -43,7 +43,8 @@ module "preservica_config_queue" {
     queue_name = "${local.environment}-preservica-config"
     topic_arn  = module.preservica_config_sns.sns_arn
   })
-  kms_key_id = module.dr2_developer_key.kms_key_arn
+  kms_key_id         = module.dr2_developer_key.kms_key_arn
+  visibility_timeout = 60
 }
 
 module "preservica_config_lambda" {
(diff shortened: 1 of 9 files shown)


deploy_preservica_config.tf
```diff
@@ -0,0 +1,80 @@
...
+    Name = "Preservica Config SNS"
+  }
+  topic_name  = "${local.environment}-preservica-config"
+  kms_key_arn = module.dr2_developer_key.kms_key_arn
+  sqs_subscriptions = {
+    preservica_config_queue = module.preservica_config_queue.sqs_arn
+  }
+}
+
+module "preservica_config_queue" {
```
kurtismash, line 38: May require a greater than default max receives as the config may have some decencies on other invocations of this lambda. The default of 3 seems quite low as well.

MancunianSam: How many do you reckon? 6 seems ok. I'll put a note in to change the default if that's sensible.

templates/s3/preservica_config_bucket_policy.json.tpl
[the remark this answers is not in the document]
MancunianSam: It doesn't. You get implicit access to the bucket from the account. You can set a deny policy with `NotPrincipal` or string not equals the role. Do we need that though, this isn't massively sensitive and the lambda IAM policy restricts what it can access. This is mostly just belt and braces.

deploy_preservica_config.tf
```diff
@@ -0,0 +1,80 @@
...
+  queue_name = "${local.environment}-preservica-config"
+  sqs_policy = templatefile("${path.module}/templates/sqs/sns_send_message_policy.json.tpl", {
+    account_id = data.aws_caller_identity.current.account_id, queue_name = "${local.environment}-preservica-config"
+    queue_name = "${local.environment}-preservica-config"
+    topic_arn  = module.preservica_config_sns.sns_arn
+  })
+  kms_key_id = module.dr2_developer_key.kms_key_arn
+}
+
+module "preservica_config_lambda" {
```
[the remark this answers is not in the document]
MancunianSam: I've added in a KMS key to encrypt th elog group.

deploy_preservica_config.tf
```diff
@@ -0,0 +1,82 @@
...
+module "preservica_config_queue" {
+  source     = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
+  queue_name = "${local.environment}-preservica-config"
+  sqs_policy = templatefile("${path.module}/templates/sqs/sns_send_message_policy.json.tpl", {
+    account_id = data.aws_caller_identity.current.account_id, queue_name = "${local.environment}-preservica-config"
+    queue_name = "${local.environment}-preservica-config"
+    topic_arn  = module.preservica_config_sns.sns_arn
+  })
+  kms_key_id         = module.dr2_developer_key.kms_key_arn
+  visibility_timeout = 60
```
kurtismash, line 47: > To allow your function time to process each batch of records, set the source queue's [visibility timeout](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html) to at least six times the [timeout that you configure](https://docs.aws.amazon.com/lambda/latest/dg/configuration-function-common.html#configuration-common-summary) on your function.

https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#events-sqs-queueconfig:~:text=To%20allow%20your%20function%20time%20to%20process%20each%20batch%20of%20records%2C%20set%20the%20source%20queue%27s%20visibility%20timeout%20to%20at%20least%20six%20times%20the%20timeout%20that%20you%20configure%20on%20your%20function.

deploy_preservica_config.tf
```diff
@@ -0,0 +1,82 @@
...
+  }
+  memory_size = 512
+  runtime     = "java17"
+  vpc_config = {
+    subnet_ids         = module.vpc.private_subnets
+    security_group_ids = [module.outbound_https_access_only.security_group_id]
+  }
+  plaintext_env_vars = {
+    SECRET_NAME    = aws_secretsmanager_secret.preservica_secret.name
+    PRESERVICA_URL = data.aws_ssm_parameter.preservica_url.value
```
kurtismash, line 76: `PRESERVICA_SECRET_NAME` and `PRESERVICA_API_URL` to match #11?

deploy_preservica_config.tf
MancunianSam: Fair enough, I've set it to 360.

kurtismash (approved):

PR #14, opened 2023-08-01
Add initial last polled entry.

MancunianSam:
This is a one off change that will be overwritten by the entity event
generator. The item changes will be ignored by terraform.

This allows us to keep the dynamo client as it currently is where there
is an error if the key is missing.

I've also increased the memory as 256Mb is a bit low for Java lambdas
and put it into the VPC so we can add the IP for the nat instance to the
Preservica security group/hosted preservica IP list.


templates/dynamo/initial_last_updated_item.json.tpl
```diff
@@ -0,0 +1,8 @@
+{
+  "id": {
+    "S": "LastPolled"
+  },
+  "datetime": {
+    "S": "1970-01-01T00:00Z"
```
techncl, line 6: Is this format ok? I thought it had to be same as the Preservica one (with the milliseconds and offset) e.g.
`2023-04-01T00:00:00.000000+0000`

entity_event_generator.tf
```diff
@@ -45,6 +50,15 @@ module "entity_event_lambda_updated_since_query_start_datetime_table" {
   table_name    = local.last_polled_table_name
 }
 
+resource "aws_dynamodb_table_item" "initial_start_date" {
```
techncl, line 53: Maybe `initial_start_datetime`?

techncl (commented):
Looks good! Just 2 comments.

templates/dynamo/initial_last_updated_item.json.tpl
MancunianSam: You're using `OffsetDateTime` in the code which is working with Preservica and this parses into an `OffsetDateTime` I've tried it and it's working ok. Once the event generator runs once, it'll be overwritten anyway.

entity_event_generator.tf
```diff
@@ -50,7 +50,7 @@ module "entity_event_lambda_updated_since_query_start_datetime_table" {
   table_name    = local.last_polled_table_name
 }
 
-resource "aws_dynamodb_table_item" "initial_start_date" {
+resource "aws_dynamodb_table_item" "initial_start_date_time" {
```
techncl, line 53: Shouldn't it be`initial_start_datetime`?

MancunianSam: Ah sorry, I can't read. I've updated it.

techncl (approved):

PR #16, opened 2023-08-02
DR2-1149 feat: ingest-parsed-court-document-event-handler-lambda terraform

MancunianSam:
There are a few changes lumped in here:

* The lambda and SQS queue for the ingest-event-handler lambda. These
  are in a seperate tf file.
* The raw cache bucket and the skeleton step function are in the common
  folder.
* The disaster recovery lambda policy was the same as the one needed
  here apart from variable names so I've renamed it and used it for
  both.
* The main KMS key now allows the admin role to decrypt on integration
  to make debugging there easier.


common.tf
```diff
@@ -1,5 +1,8 @@
 locals {
-  az_count = local.environment == "prod" ? 2 : 1
+  az_count                      = local.environment == "prod" ? 2 : 1
+  ingest_raw_cache_bucket_name  = "${local.environment}-dr2-ingest-raw-cache"
+  pre_ingest_step_function_name = "${local.environment_title}PreIngestStepFunction"
```
kurtismash, line 4: We need to agree some naming here. Preservica call `preingest` the process of forming a SIP/OPEX package, validating metadata, etc. Our Step Function in effect is doing both `preingest` and `ingest` (the Preservica bit), and may extend to post-ingest where we wait for the DR copy to be written. There's also the possibility of more complex operations upstream of this step function, e.g. validating a TDR BagIt or mapping complex assets together. I err'ed on the side of caution and called this the `ingest` step function which leaves `preingest` to the upstream process that takes a package and puts it into our DR2 BagIt format, but that's still different to Preservica's terminology.

ingest_parsed_court_document_event_handler.tf
```diff
@@ -0,0 +1,47 @@
...
+  ingest_parsed_court_document_event_handler_lambda_name = "${local.environment}-ingest-parsed-court-document-event-handler"
+}
+module "ingest_parsed_court_document_event_handler_sqs" {
+  source     = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
+  queue_name = local.ingest_parsed_court_document_event_handler_queue_name
+  sqs_policy = templatefile("./templates/sqs/sqs_access_policy.json.tpl", {
+    account_id = var.dp_account_number, //TODO Restrict this to the SNS topic ARN when it's created
+    queue_name = local.ingest_parsed_court_document_event_handler_queue_name
+  })
+  redrive_maximum_receives = 3
```
kurtismash, line 12: Standard SQS-Lambda complaint - max receives >= 5 and visibility timeout >= 6*lambda timeout.
...or use the Lambda Event Mapping scaling config.

ingest_parsed_court_document_event_handler.tf
```diff
@@ -0,0 +1,47 @@
...
+      lambda_name                                          = local.ingest_parsed_court_document_event_handler_lambda_name
+      step_function_arn                                    = module.pre_ingest_step_function.step_function_arn
+    })
+  }
+  memory_size = 512
+  runtime     = "java17"
+  vpc_config = {
+    subnet_ids         = module.vpc.private_subnets
+    security_group_ids = [module.outbound_https_access_only.security_group_id]
+  }
```
kurtismash, line 38: This Lambda doesn't need to be in the VPC.

templates/iam_policy/ingest_parsed_court_document_event_handler_lambda_policy.json.tpl
```diff
@@ -0,0 +1,48 @@
...
+      "Action": [
+        "s3:PutObject*",
+        "s3:GetObject"
+      ],
+      "Effect": "Allow",
+      "Resource": [
+        "arn:aws:s3:::${bucket_name}",
+        "arn:aws:s3:::${bucket_name}/*"
+      ],
+      "Sid": "readWriteDisasterRecoveryBucket"
```
kurtismash, line 31: Sid

ingest_parsed_court_document_event_handler.tf
MancunianSam: I remember thinking about this, then I forgot. I've changed this to 5. I've increased the lambda timeout to 30 seconds because 15 is cutting it close on a cold start and I've set the queue visibility timeout to 180.

ingest_parsed_court_document_event_handler.tf
MancunianSam: Force of habit, I've taken it out.

templates/iam_policy/ingest_parsed_court_document_event_handler_lambda_policy.json.tpl
MancunianSam: Copy and paste is fun, I've changed it.

kurtismash (dismissed):

common.tf
```diff
@@ -1,7 +1,7 @@
 locals {
   az_count                      = local.environment == "prod" ? 2 : 1
   ingest_raw_cache_bucket_name  = "${local.environment}-dr2-ingest-raw-cache"
-  pre_ingest_step_function_name = "${local.environment_title}PreIngestStepFunction"
+  pre_ingest_step_function_name = "${local.environment_title}IngestStepFunction"
```
techncl, line 4: Is it fine that there is no hyphen after "environment_title"?

kurtismash:
Lambda will also need read on an upstream bucket, do we want to create a mock for this and add it now?

common.tf
MancunianSam: It's title case so it comes out as IntgIngestStepFunction which is ok I think.

kurtismash: Why is this name title case but all the others are hyphen?

MancunianSam:
I've added in a test bucket for us to use.
I've re-used the `lambda_access_bucket_policy` which has `PutObject` permissions which this doesn't need but I didn't want to create a new template for a test bucket. Hopefully we'll come up with a better solution for e2e testing at some point.

common.tf
MancunianSam: Because I don't like consistency? Come to think of it, the role has a hyphen in it as well. Ok, I'll change it.

PR #17, opened 2023-08-14
Add ingest mapper lambda.

MancunianSam:
I've also added in the dynamo table which technically belongs under
https://national-archives.atlassian.net/browse/DR2-1145 but I can't test
this lambda without that.


techncl (approved):
Looks good!

PR #19, opened 2023-08-16
Add ingest raw staging bucket.

MancunianSam:
This bucket will probably have more than one lambda accessing it so I've
changed the lambda bucket policy template to accept an array instead of
a single lambda. I've updated the other buckets that use this template.

This has no effect at the moment because the account has access to the
bucket by default but if we want to change the policy to a deny unless
style one later, this will help.

The raw cache bucket has already been created under another PR.


techncl (approved):
Looks good!

techncl (approved):

PR #20, opened 2023-08-29
Add terraform lambda to deploy policy.

MancunianSam:
I've also changed the policy to take an array so I don't have to update
the json template every time there is a new lambda.

### DIFF
--- a/deploy_roles.tf
+++ b/deploy_roles.tf
@@ -12,9 +12,12 @@ module "deploy_lambda_policy" {
   source = "git::https://github.com/nationalarchives/da-terraform-modules//iam_policy"
   name   = "${local.environment_title}DPGithubActionsDeployLambdaPolicy"
   policy_string = templatefile("${path.module}/templates/iam_policy/deploy_lambda_policy.json.tpl", {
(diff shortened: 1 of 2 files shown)


deploy_roles.tf
```diff
@@ -12,9 +12,12 @@ module "deploy_lambda_policy" {
...
-    slack_notification_lambda_arn      = module.slack_notifications_lambda.lambda_arn
-    entity_event_generation_lambda_arn = module.entity_event_generator_lambda.lambda_arn
-    bucket_name                        = "mgmt-dp-code-deploy"
+    lambda_arns = jsonencode([
+      module.download_metadata_and_files_lambda.lambda_arn,
+      module.slack_notifications_lambda.lambda_arn,
+      module.entity_event_generator_lambda.lambda_arn,
+      module.ingest_mapper_lambda.lambda_arn
+    ])
+    bucket_name = "mgmt-dp-code-deploy"
```
techncl, line 21: This isn't a part of your changes but is it fine that the bucket is "dp" and not "dr2"?

techncl (approved):

PR #21, opened 2023-08-30
DR2-1165 Terraform for ingest-asset-opex-creator lambda

MancunianSam:
* Add the lambda
* Add a policy which allows
  * Read from raw bucket
  * Write to staging bucket
  * Cloudwatch permissions
  * Batch get permissions on the dynamo table
  * Query permissions on the GSI
* Add the lambda role arn to the KMS policy

### DIFF
--- a/common.tf
+++ b/common.tf
@@ -1,10 +1,11 @@
 locals {
-  az_count                         = local.environment == "prod" ? 2 : 1
-  ingest_raw_cache_bucket_name     = "${local.environment}-dr2-ingest-raw-cache"
-  ingest_staging_cache_bucket_name = "${local.environment}-dr2-ingest-staging-cache"
-  pre_ingest_step_function_name    = "${local.environment_title}-ingest-step-function"
-  additional_user_roles            = local.environment == "intg" ? [data.aws_ssm_parameter.dev_admin_role.value] : []
-  files_dynamo_table_name          = "${local.environment}-dr2-files"
+  az_count                                = local.environment == "prod" ? 2 : 1
+  ingest_raw_cache_bucket_name            = "${local.environment}-dr2-ingest-raw-cache"
+  ingest_staging_cache_bucket_name        = "${local.environment}-dr2-ingest-staging-cache"
+  pre_ingest_step_function_name           = "${local.environment_title}-ingest-step-function"
+  additional_user_roles                   = local.environment == "intg" ? [data.aws_ssm_parameter.dev_admin_role.value] : []
+  files_dynamo_table_name                 = "${local.environment}-dr2-files"
+  files_table_global_secondary_index_name = "BatchParentPathIdx"
 }
 resource "random_password" "preservica_password" {
   length = 20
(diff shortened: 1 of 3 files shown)


ingest_asset_opex_creator.tf
```diff
@@ -0,0 +1,32 @@
+locals {
+  ingest_asset_opex_creator_lambda_name = "${local.environment}-ingest-mapper-opex-creator"
```
kurtismash, line 2: `mapper`??

common.tf
```diff
@@ -1,10 +1,11 @@
...
-  ingest_staging_cache_bucket_name = "${local.environment}-dr2-ingest-staging-cache"
-  pre_ingest_step_function_name    = "${local.environment_title}-ingest-step-function"
-  additional_user_roles            = local.environment == "intg" ? [data.aws_ssm_parameter.dev_admin_role.value] : []
-  files_dynamo_table_name          = "${local.environment}-dr2-files"
+  az_count                                = local.environment == "prod" ? 2 : 1
+  ingest_raw_cache_bucket_name            = "${local.environment}-dr2-ingest-raw-cache"
+  ingest_staging_cache_bucket_name        = "${local.environment}-dr2-ingest-staging-cache"
+  pre_ingest_step_function_name           = "${local.environment_title}-ingest-step-function"
+  additional_user_roles                   = local.environment == "intg" ? [data.aws_ssm_parameter.dev_admin_role.value] : []
+  files_dynamo_table_name                 = "${local.environment}-dr2-files"
```
kurtismash, line 7: Is `files` the best name for this table as it also contains folders and assets?

ingest_asset_opex_creator.tf
```diff
@@ -0,0 +1,32 @@
...
+      gsi_name                = local.files_table_global_secondary_index_name
+
+    })
+  }
+  memory_size = 512
+  runtime     = "java17"
+  plaintext_env_vars = {
+    DYNAMO_TABLE_NAME  = local.files_dynamo_table_name
+    DYNAMO_GSI_NAME    = local.files_table_global_secondary_index_name
+    SOURCE_BUCKET      = local.ingest_raw_cache_bucket_name
```
kurtismash, line 25: Would `SOURCE_BUCKET` be better to be passed in the event as the prefix will also be required from the event anyway?

templates/iam_policy/ingest_asset_opex_creator_policy.json.tpl
```diff
@@ -0,0 +1,54 @@
...
+      "Resource": [
+        "${dynamo_db_arn}",
+        "${dynamo_db_arn}/index/${gsi_name}"
+      ],
+      "Sid": "getAndQueryDynamoDB"
+    },
+    {
+      "Action": [
+        "s3:GetObject",
+        "s3:ListBucket"
```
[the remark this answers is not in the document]
MancunianSam: You don't need either `ListBucket` permission. I read that you did but that might be for doing it through the console. I've taken them both out.

common.tf
MancunianSam: https://en.wikipedia.org/wiki/Everything_is_a_file

ingest_asset_opex_creator.tf
MancunianSam: Oops, copy paste error.

ingest_asset_opex_creator.tf
MancunianSam: We're passing in batch ID and exection name to the asset-creator lambda but that's fine because the batch ID is available on the global input object, as is the execution name so we can add those to the input in the step function but the source bucket will have to be passed in from the upsert lambda which will end up looking like the json above.

kurtismash: Could we not do some Step Functions stuff here with the itemSelector field, like in https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#inline-map-state-example-params ? Our state data will contain an S3 prefix for the batch, which we'd need in each invocation as an ID alone won't be enough to figure out where the object is in S3.

kurtismash (approved):

PR #23, opened 2023-09-05
DR2-1166 Folder opex creator

MancunianSam:
Add the terraform for the folder opex lambda
Encrypt the dynamo table. The existing lambdas should work.


techncl (approved):
Looks good!

PR #24, opened 2023-09-05
Add upsert archive folder Lambda's terraform

techncl:


common.tf
```diff
@@ -92,7 +92,8 @@ module "dr2_kms_key" {
       module.slack_notifications_lambda.lambda_role_arn,
       module.ingest_mapper_lambda.lambda_role_arn,
       module.ingest_asset_opex_creator_lambda.lambda_role_arn,
-      module.ingest_folder_opex_creator_lambda.lambda_role_arn
+      module.ingest_folder_opex_creator_lambda.lambda_role_arn,
+      module.ingest_upsert_archives_folder_lambda
```
MancunianSam, line 96: You need to add `.lambda_role_arn` to the end here. That's why the tests are failing I think.

techncl: Ah, 🤦 . Thanks

PR #26, opened 2023-09-08
DR2 1202 add event bridge alarm and slack events

MancunianSam:
Add eventbridge rules.
There are three rules.

One for the state change on our cloudwatch alarms when it moves to ALARM
One for the state change on our cloudwatch alarms when it moves to OK

The reason we need to is so we can have different slack icons for each.

Another one for sending messages directly to Slack.

I've removed the existing notifications and updated the queues to
remove the sns topic variable.
I've also deleted the notifications lambda.


common.tf
```diff
@@ -168,3 +170,59 @@ module "files_table" {
...
+}
+
+module "dev_slack_message_eventbridge_rule" {
+  source              = "git::https://github.com/nationalarchives/da-terraform-modules//eventbridge_api_destination_rule"
+  api_destination_arn = module.eventbridge_alarm_notifications_destination.api_destination_arn
+  event_pattern       = templatefile("${path.module}/templates/eventbridge/custom_detail_type_event_pattern.json.tpl", { detail_type = "DR2DevMessage" })
+  name                = "${local.environment}-eventbridge-dev-slack-message"
+  input_transformer = {
+    input_paths = {
+      "message" = "$.detail.message"
```
techncl, line 221: Should we rename either this "message" or the one on line 225, to "slackMessage"?

common.tf
```diff
@@ -168,3 +170,59 @@ module "files_table" {
     }
   ]
 }
+
+data "aws_ssm_parameter" "slack_token" {
+  name            = "/mgmt/slack/token"
```
techncl, line 175: Should there be some sort of `${local.environment}` here instead of `mgmt` or is this fine?

common.tf
```diff
@@ -168,3 +170,59 @@ module "files_table" {
...
+  with_decryption = true
+}
+
+data "aws_ssm_parameter" "dr2_notifications_slack_channel" {
+  name = "/mgmt/slack/notifications/channel"
```
techncl, line 180: Should there be some sort of `${local.environment}` here instead of `mgmt` or is this fine?

common.tf
```diff
@@ -104,9 +103,12 @@ module "dr2_developer_key" {
   source   = "git::https://github.com/nationalarchives/da-terraform-modules//kms"
   key_name = "${local.environment}-kms-dr2-dev"
   default_policy_variables = {
-    user_roles    = [data.aws_ssm_parameter.dev_admin_role.value, module.preservica_config_lambda.lambda_role_arn]
+    user_roles = [
+      data.aws_ssm_parameter.dev_admin_role.value,
+      module.preservica_config_lambda.lambda_role_arn,
```
techncl, line 108: There is an extra comma here, is that deliberate?

techncl (commented):
Looks good; just a few minor comments.

common.tf
MancunianSam: We can change this one and the one on line 255. I've changed them both.

common.tf
MancunianSam: Same as above

common.tf
MancunianSam: No, it's valid but trailing commas are annoying. I've removed it.

techncl (approved):
👍

PR #30, opened 2023-09-15
DR2 1217 add put events policy

MancunianSam:
Add put events to upsert lambda.
Also update the da-configurations hash. The changes are from other teams so don't affect us but it's nice to keep it up to date.


techncl (approved):
Looks good.

PR #32, opened 2023-09-20
Update the IP lock lambda to run once per hour.

MancunianSam:
The IP locking is in place so this won't be so annoying now.


techncl (approved):
Looks good!

PR #33, opened 2023-09-20
Use environment specific terraform roles

MancunianSam:
The terraform role was hard coded for integration.

### DIFF
--- a/common.tf
+++ b/common.tf
@@ -94,7 +94,7 @@ module "dr2_kms_key" {
       module.ingest_folder_opex_creator_lambda.lambda_role_arn,
       module.ingest_upsert_archives_folder_lambda.lambda_role_arn
     ], local.additional_user_roles)
-    ci_roles      = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/${title(local.environment)}TerraformRole"]
+    ci_roles      = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/${local.environment_title}TerraformRole"]
     service_names = ["cloudwatch", "sns"]
   }
 }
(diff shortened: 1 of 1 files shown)


common.tf
```diff
@@ -94,7 +94,7 @@ module "dr2_kms_key" {
       module.ingest_folder_opex_creator_lambda.lambda_role_arn,
       module.ingest_upsert_archives_folder_lambda.lambda_role_arn
     ], local.additional_user_roles)
-    ci_roles      = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/IntgTerraformRole"]
+    ci_roles      = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/${title(local.environment)}TerraformRole"]
```
techncl, line 97: Couldn't you use `local.environment_title` here instead?

common.tf
```diff
@@ -107,7 +107,7 @@ module "dr2_developer_key" {
       data.aws_ssm_parameter.dev_admin_role.value,
       module.preservica_config_lambda.lambda_role_arn
     ]
-    ci_roles      = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/IntgTerraformRole"]
+    ci_roles      = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/${title(local.environment)}TerraformRole"]
```
techncl, line 110: Couldn't you use `local.environment_title` here instead?

techncl (commented):
Looks good! Just a minor comment

common.tf
MancunianSam: Good spot, I forgot about that. I've changed it. Same for the one below.

techncl (approved):

PR #39, opened 2023-10-06
Update AWS provider and lambda runtime

MancunianSam:
We were sat on 5.1.0 and there is a 5.19 version of the AWS provider.
This was stopping us using python3.11 for lambdas.
This works now


techncl (approved):
Looks good!

PR #40, opened 2023-10-10
Small fixes to allow testing on staging.

MancunianSam:
The first is to allow the admin role to use the KMS key which lets us
put and get stuff from the test input bucket.

The second is to shorten the name of the bucket because
staging-ingest-parsed-court-document-event-handler-test-input-logs is
too long. We may not need individual log buckets in the future but for
now, this will do.


techncl (approved):
Looks good!

PR #41, opened 2023-10-11
DR2-1245 create sample files bucket

MancunianSam:
- Small fixes to allow testing on staging.
- Add sample files bucket

### DIFF
--- a/common.tf
+++ b/common.tf
@@ -1,6 +1,7 @@
 locals {
   az_count                                = local.environment == "prod" ? 2 : 1
   ingest_raw_cache_bucket_name            = "${local.environment}-dr2-ingest-raw-cache"
+  sample_files_bucket_name                = "${local.environment}-dr2-sample-files"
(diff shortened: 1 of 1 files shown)


common.tf
```diff
@@ -129,6 +130,13 @@ module "ingest_raw_cache_bucket" {
   kms_key_arn = module.dr2_kms_key.kms_key_arn
 }
 
+module "sample_files_bucket" {
```
techncl, line 133: Event though this is for sample files, should we add a bucket policy?

techncl (commented):
Looks good! Just a minor comment

common.tf
MancunianSam: It's got one. I updated the [da-terraform-modules](https://github.com/nationalarchives/da-terraform-modules/blob/main/s3/templates/default_bucket_policy.json.tpl) to add a default bucket policy if there isn't one set. It just makes sure that all requests are over SSL.

techncl (approved):

PR #43, opened 2023-10-18
Add startWorkflow lambda terraform

techncl:


common.tf
```diff
@@ -93,7 +93,8 @@ module "dr2_kms_key" {
       module.ingest_mapper_lambda.lambda_role_arn,
       module.ingest_asset_opex_creator_lambda.lambda_role_arn,
       module.ingest_folder_opex_creator_lambda.lambda_role_arn,
-      module.ingest_upsert_archive_folders_lambda.lambda_role_arn
+      module.ingest_upsert_archive_folders_lambda.lambda_role_arn,
+      module.ingest_start_workflow_lambda.lambda_role_arn
```
MancunianSam, line 97: You don't need to add the lambda to the KMS key. This is  only needed for the lambdas which access S3 or are triggered from encrypted queues.

common.tf
```diff
@@ -175,6 +177,7 @@ module "ingest_step_function_policy" {
     ingest_upsert_archive_folders_lambda_name = local.ingest_upsert_archive_folders_lambda_name
     ingest_asset_opex_creator_lambda_name     = local.ingest_asset_opex_creator_lambda_name
     ingest_folder_opex_creator_lambda_name    = local.ingest_folder_opex_creator_lambda_name
+    ingest_start_workflow_lambda_name         = local.ingest_start_workflow_lambda_name
```
MancunianSam, line 180: You need to add a line into `ingest_step_function_policy.json.tpl` to use this variable.

MancunianSam (commented):
Looks good. Just a couple of small changes

PR #44, opened 2023-10-23
Add dollar sign

MancunianSam:
Without the dollar sign it was showing {account_id} rather than the
account id. This should work now.


techncl (approved):
🤦 sorry about that. Looks good!

PR #45, opened 2023-10-23
DR2-1240 DR2-1302 Add NACL rules and use NAT gateway

MancunianSam:


kurtismash:
Not sure why we need 443 rules but I've been unable to figure it out whilst timeboxed. Don't think it's a showstopper as we have nothing that could be connected to in our VPC - yay lambdas! Will put a ticket in the backlog to log deeper into this.

kurtismash (approved):

PR #48, opened 2023-11-03
Add terraform for Parent folder Opex creator Lambda

techncl:


ingest_parent_folder_opex_creator.tf
```diff
@@ -0,0 +1,29 @@
...
+      account_id  = var.account_number
+      lambda_name = local.ingest_parent_folder_opex_creator_lambda_name
+    })
+  }
+  memory_size = 512
+  runtime     = "java17"
+  plaintext_env_vars = {
+    STAGING_CACHE_BUCKET = local.ingest_staging_cache_bucket_name
+  }
+  vpc_config = {
```
MancunianSam, line 21: This isn't talking to Preservica so it doesn't need to be in the VPC so you can take lines 21-24 out.

templates/s3/lambda_access_bucket_policy.json.tpl
```diff
@@ -8,7 +8,8 @@
       },
       "Action": [
         "s3:GetObject",
-        "s3:PutObject"
+        "s3:PutObject",
+        "s3:ListBucket"
       ],
       "Resource": "arn:aws:s3:::${bucket_name}/*"
```
MancunianSam, line 14: You need to add a resource for the bucket as well as the objects. `"arn:aws:s3:::${bucket_name}"`

ingest_parent_folder_opex_creator.tf
```diff
@@ -0,0 +1,29 @@
...
+    subnet_ids         = module.vpc.private_subnets
+    security_group_ids = [module.outbound_https_access_only.security_group_id]
+  }
+  tags = {
+    Name      = local.ingest_parent_folder_opex_creator_lambda_name
+    CreatedBy = "dp-terraform-environments"
```
MancunianSam, line 27: CreatedBy is set at the provider level in the terraform so you don't need it here.

MancunianSam (commented):
Looks good. Couple of minor things. You also need to add the lambda name in here https://github.com/nationalarchives/dr2-terraform-environments/blob/main/deploy_roles.tf#L24 so Github can deploy it.

MancunianSam (approved):

PR #51, opened 2023-11-21
DR2-1313 Add Preservica copy role and policy

MancunianSam:


tna_to_preservica.tf
techncl, line 1: Should this have the word "role" in it or is it possible that we add more than roles in the future?

techncl (commented):
Looks good! Just some minor comments.

PR #59, opened 2023-12-01
Add anonymiser lambda terraform

MancunianSam:
There is the lambda, a queue to trigger the lambda and an SNS topic
subscription to the TRE out topic for the SQS queue.


anonymiser.tf
```diff
@@ -0,0 +1,54 @@
...
+    court_document_package_anonymiser_queue = module.court_document_package_anonymiser_sqs.sqs_arn
+  }
+  policies = {
+    "${local.court_document_anonymiser_lambda_name}-policy" = templatefile("./templates/iam_policy/anonymiser_lambda_policy.json.tpl", {
+      anonymiser_test_input_queue         = module.court_document_package_anonymiser_sqs.sqs_arn
+      ingest_court_document_handler_queue = module.ingest_parsed_court_document_event_handler_sqs.sqs_arn
+      output_bucket_name                  = local.ingest_parsed_court_document_event_handler_test_bucket_name
+      account_id                          = var.account_number
+      lambda_name                         = local.court_document_anonymiser_lambda_name
+      tre_bucket_arn                      = module.tre_config.terraform_config["prod"]["s3_court_document_pack_out_arn"]
```
techncl, line 20: 1. Is it fine that we have "prod" hardcoded? Could it not be `${local.environment}`? If not, could "prod" be saved to a variable?

anonymiser.tf
```diff
@@ -0,0 +1,54 @@
...
+      tre_kms_arn                         = module.tre_config.terraform_config["prod_s3_court_document_pack_out_kms_arn"]
```
techncl, line 21: Same question as above and same question for line 39 and 50

common.tf
```diff
@@ -89,9 +89,15 @@ module "dr2_kms_key" {
       module.e2e_tests_ecs_task_role.role_arn,
       module.copy_tna_to_preservica_role.role_arn,
       "arn:aws:iam::${module.tre_config.account_numbers["prod"]}:role/prod-tre-editorial-judgment-out-copier",
```
techncl, line 91: 1. Could this be saved to a variable?
2. Is it fine that we are using the hardcoded "prod" in both places?

techncl (commented):
Looks good, just some minor comments

anonymiser.tf
MancunianSam: We need "prod" hard coded because we're taking judgments from TRE prod to our intg environment. I've moved `module.tre_config.terraform_config["prod"]` into a local variable which has tidied things up a bit.

anonymiser.tf
MancunianSam: I'm going to leave this one on line 21. It's different to the line above because of the way TRE have done their config so there's not much point in moving it to a variable.

techncl (approved):
👍

PR #64, opened 2023-12-12
DR2-1363 Add eventbridge rule for security hub.

MancunianSam:
This will trigger when findings are imported or updated in Security Hub.

We'll have to see if this works and whether it's too chatty to be any
user.


MancunianSam:
The security hub rule is far too chatty and there doesn't seem to be a way to filter existing things so I'll close this and we'll re-enable it when we've fixed some of the findings.

PR #68, opened 2024-01-18
DR2-1417 Update lambda policy.

MancunianSam:
Because this is cross account access, the lambda policy needs GetObject
permission on the TRE bucket and decrypt permissions on the KMS key from
TRE for this to work.

### DIFF
--- a/anonymiser.tf
+++ b/anonymiser.tf
@@ -4,17 +4,16 @@ locals {
   tre_terraform_prod_config             = module.tre_config.terraform_config["prod"]
 }
 module "court_document_package_anonymiser_lambda" {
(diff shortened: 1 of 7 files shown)


techncl (approved):
Looks good! I'm guessing there probably isn't a way to avoid the Iam policy duplication (e.g. a template where you pass in a json object)?

MancunianSam:
There sort of is, we could have one template that just has the statements and then have that as a variable in the json file but that would be much harder to read than just having two. This way, it's obvious that there's separate policies.

techncl (approved):

techncl (approved):

PR #86, opened 2024-03-04
Add Preservica version Lambda and resources

techncl:


templates/iam_policy/get_latest_preservica_version_lambda_policy.json.tpl
```diff
@@ -0,0 +1,38 @@
+{
+  "Statement": [
+    {
+      "Action": [
+        "dynamodb:BatchGetItem"
+      ],
+      "Effect": "Allow",
+      "Resource": "${dynamo_db_arn}",
+      "Sid": "readWriteDynamoDB"
```
MancunianSam, line 9: The permissions are only for read so can you change the Sid to `readDynamoDb` please.

MancunianSam (commented):
Looks good, just one minor comment.

get_latest_preservica_version.tf
```diff
@@ -0,0 +1,70 @@
...
+    })
+  }
+  timeout_seconds = 180
+  memory_size     = local.java_lambda_memory_size
+  runtime         = local.java_runtime
+  tags            = {}
+  lambda_invoke_permissions = {
+    "events.amazonaws.com" = module.get_latest_preservica_version_cloudwatch_event.event_arn
+  }
+  vpc_config = {
```
kurtismash, line 35: As we'll be running this Lambda against Preservica's demo/reltest system it doesn't need to be in our VPC.

templates/iam_policy/get_latest_preservica_version_lambda_policy.json.tpl
kurtismash: What do we want to happen when Preservica update their demo/reltest system? If we're only writing to ddb this Lambda will continue to generate events on a new version until someone changes the value in ddb.

MancunianSam: Isn't that the idea, when they update their test system, we manually update the value in Dynamo. That seems safer than getting the lambda to do it because we might miss the initial message.

get_latest_preservica_version.tf
```diff
@@ -0,0 +1,70 @@
...
+  item       = templatefile("${path.module}/templates/dynamo/dr2_preservica_version.json.tpl", {})
+  table_name = local.dr2_preservica_version_table_name
+  lifecycle {
+    ignore_changes = [item]
+  }
+}
+
+module "latest_preservica_version_topic" {
+  source = "git::https://github.com/nationalarchives/da-terraform-modules//sns"
+  sns_policy = templatefile("${path.module}/templates/sns/latest_preservica_version_topic_policy.json.tpl", {
```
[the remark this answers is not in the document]
MancunianSam: That was for S3 I think that doesn't need two way permissions. This makes sense I think in that the topic is restricted just to this lambda.

kurtismash:
@MancunianSam I'll leave this to you to approve.

MancunianSam (approved):
:+1:

PR #88, opened 2024-03-07
DR2-1484 Guard duty alerts

MancunianSam:
Add Guard Duty finding alerts.

I can't seem to use the existing slack template as terraform gets very
upset when I try to use newlines inside a json template.

I've also hard coded the prod dev channel in the slack template so we
don't lose any findings in the other noise.

### DIFF
--- a/common.tf
+++ b/common.tf
@@ -340,10 +340,10 @@ module "guard_duty_findings_eventbridge_rule" {
   api_destination_arn = module.eventbridge_alarm_notifications_destination.api_destination_arn
   api_destination_input_transformer = {
     input_paths = {
-      "account": "$.account",
-      "id": "$.detail.id",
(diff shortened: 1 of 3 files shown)


kurtismash:
The existing alerts for Guard Duty have been [configured though Cloud Custodian](https://github.com/nationalarchives/tna-custodian/blob/master/custodian/policies/guardduty/guard-duty-notify.yml). I'll approve this, it's up to you if you want to do it through native EventBridge or CC.

Do you think it's worth pulling the finding type out to the Slack message?

MancunianSam:
I'll leave it like this as a native EventBridge rule for now. If we end up keeping Cloud Custodian long term I might shift it over. The slack message has the link to the finding and the account number so I think that'll be enough.

PR #90, opened 2024-03-18
DR2-1549 Update to allow eventbridge publish

MancunianSam:
Also, addsome missing lambdas to the deploy policy temporarily and add
the ListBucket permission for the code deploy role.


techncl (approved):
Looks good!

PR #93, opened 2024-04-15
DR2-1584 Create entity queue and DR user

MancunianSam:
This is the user which will be used by the disaster recovery service.

We also didn't have a terraformed entity event queue listening to the
topic so I've added this in.


techncl (approved):
Looks good!

PR #94, opened 2024-04-18
DR2-1589 Add dr2 to names

MancunianSam:
So that we don't have to delete and recreate queues and step functions
and potentially miss messages. We're going to do this in two stages.
This is the first stage where we duplicate the queues, lambdas and step
function.
The latest code will then be deployed to the new lambda functions.
Then we'll disable the trigger on the parsed court document lambda and
move the subscription to the new topic.
Finally, the old queues, lambdas and step functions can be deleted.


common.tf
```diff
@@ -99,17 +100,24 @@ module "dr2_kms_key" {
...
+
+      module.dr2_ingest_parsed_court_document_event_handler_lambda.lambda_role_arn,
+      module.dr2_ingest_mapper_lambda.lambda_role_arn,
+      module.dr2_ingest_asset_opex_creator_lambda.lambda_role_arn,
+      module.dr2_ingest_folder_opex_creator_lambda.lambda_role_arn,
+      module.dr2_ingest_upsert_archive_folders_lambda.lambda_role_arn,
+      module.dr2_ingest_parent_folder_opex_creator_lambda.lambda_role_arn,
+      module.dr2_ingest_asset_reconciler_lambda.lambda_role_arn,
+
       module.e2e_tests_ecs_task_role.role_arn,
```
techncl, line 118: Is it ok that these don't start with `dr2`?

deploy_preservica_config.tf
```diff
@@ -31,6 +32,23 @@ module "preservica_config_sns" {
   }
 }
 
+module "dr2_preservica_config_sns" {
+  source = "git::https://github.com/nationalarchives/da-terraform-modules//sns"
+  sns_policy = templatefile("${path.module}/templates/sns/s3_notifications_policy.json.tpl", {
+    topic_name  = "${local.environment}-preservica-config"
+    bucket_name = "${local.environment}-dr2-preservica-config"
```
techncl, line 39: Should this be `local.preservica_config_bucket_name`?

deploy_preservica_config.tf
```diff
@@ -31,6 +32,23 @@ module "preservica_config_sns" {
...
+    account_id  = data.aws_caller_identity.current.account_id
+  })
+  tags = {
+    Name = "Preservica Config SNS"
+  }
+  topic_name  = "${local.environment}-preservica-config"
```
techncl, line 45: Since there are 2 instances of this, should we save it to a local variable?

deploy_preservica_config.tf
```diff
@@ -31,6 +32,23 @@ module "preservica_config_sns" {
...
+  kms_key_arn = module.dr2_developer_key.kms_key_arn
+  sqs_subscriptions = {
+    preservica_config_queue = module.preservica_config_queue.sqs_arn
+  }
+}
+
 module "preservica_config_queue" {
   source     = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
   queue_name = "${local.environment}-preservica-config"
```
techncl, line 54: Since there are 2 instances of this, should we save it to a local variable?

disaster_recovery.tf
```diff
@@ -1,18 +1,6 @@
 locals {
   disaster_recovery_name = "dr2-${local.environment}-disaster-recovery"
```
techncl, line 2: I know you did not do this in this PR but is the naming of this correct?

entity_event_generator.tf
```diff
@@ -1,24 +1,60 @@
...
+      secrets_manager_secret_arn = aws_secretsmanager_secret.preservica_secret.arn
+      sns_arn                    = local.entity_event_topic_arn
+    })
+  }
+  timeout_seconds = 180
+  memory_size     = local.java_lambda_memory_size
+  runtime         = local.java_runtime
+  tags            = {}
+  lambda_invoke_permissions = {
+    "events.amazonaws.com" = module.entity_event_cloudwatch_event.event_arn
```
techncl, line 38: Should this have `dr2` in front of it?

get_latest_preservica_version.tf
```diff
@@ -57,11 +96,21 @@ resource "aws_dynamodb_table_item" "dr2_preservica_version" {
...
+module "dr2_latest_preservica_version_topic" {
   source = "git::https://github.com/nationalarchives/da-terraform-modules//sns"
   sns_policy = templatefile("${path.module}/templates/sns/latest_preservica_version_topic_policy.json.tpl", {
     lambda_role_arn = module.get_latest_preservica_version_lambda.lambda_role_arn
     sns_topic       = local.latest_preservica_version_event_topic_arn
   })
   tags       = {}
   topic_name = local.latest_preservica_version_event_topic_name
-}
+}
```
techncl, line 116: This is missing a new line

ingest_asset_opex_creator.tf
```diff
@@ -28,4 +59,4 @@ module "ingest_asset_opex_creator_lambda" {
     Name      = local.ingest_asset_opex_creator_lambda_name
     CreatedBy = "dp-terraform-environments"
   }
-}
+}
```
techncl, line 62: Missing a new line here

ingest_folder_opex_creator.tf
```diff
@@ -27,4 +56,4 @@ module "ingest_folder_opex_creator_lambda" {
     Name      = local.ingest_folder_opex_creator_lambda_name
     CreatedBy = "dp-terraform-environments"
   }
-}
+}
```
techncl, line 59: Missing a new line here

anonymiser.tf
```diff
@@ -1,10 +1,41 @@
...
+      lambda_name                         = local.court_document_anonymiser_lambda_name_old
+      tre_bucket_arn                      = local.tre_terraform_prod_config["s3_court_document_pack_out_arn"]
+      tre_kms_arn                         = module.tre_config.terraform_config["prod_s3_court_document_pack_out_kms_arn"]
+    })
+  }
+  memory_size = 128
+  runtime     = "provided.al2023"
+  plaintext_env_vars = {
+    OUTPUT_BUCKET = local.ingest_parsed_court_document_event_handler_test_bucket_name_old
+    OUTPUT_QUEUE  = module.ingest_parsed_court_document_event_handler_sqs.sqs_queue_url
```
techncl, line 33: Is it ok that this module doesn't start with `dr2`?

techncl (commented):
Looks good, just some minor comments, including:

1. Since we are renaming it, do you think it's worth changing `dr2_tna_to_preservica_copy` to `dr2_copy_tna_to_preservica` as the former name seems like it is a copy of an origina?
2. Is it ok that `closure-management.tf` is a blank file?

common.tf
MancunianSam: They're ok. The only ones I've renamed are modules for queues and lambdas where we're running two copies in parallel. I don't really want to go through and change every module name. This can just be renamed as we're not bothered about it being destroyed and recreated.

deploy_preservica_config.tf
MancunianSam: Good spot, yep.

deploy_preservica_config.tf
MancunianSam: Yeah, we can do. I've added in a variable to `preservica_config_sns_name` and `preservica_config_sns_name_old` I know they're the same as the other names but it's nice to keep things seperate.

deploy_preservica_config.tf
MancunianSam: As above, I've added in a `preservica_config_sqs_name` and `preservica_config_sqs_name_old` variable.

entity_event_generator.tf
MancunianSam: This one shouldn't but the one in the `dr2_entity_event_generator_lambda` should. I've changed it. There were a few bits missing in this file which I've changed now.

anonymiser.tf
MancunianSam: This one doesn't need to but the one on line 62 should. I've changed it.

techncl:
> Looks good, just some minor comments, including:
> 
> 1. Since we are renaming it, do you think it's worth changing `dr2_tna_to_preservica_copy` to `dr2_copy_tna_to_preservica` as the former name seems like it is a copy of an origina?
> 2. Is it ok that `closure-management.tf` is a blank file?

Sorry, I meant `dr2_copy_from_tna_to_preservica`

techncl (approved):
👍

PR #95, opened 2024-04-19
Add 'Existing IO-check' Lambda & update step function

techncl:


ingest_check_preservica_for_existing_io.tf
```diff
@@ -0,0 +1,34 @@
+locals {
+  ingest_check_preservica_for_existing_io_lambda_name = "${local.environment}-ingest-check-preservica-for-existing-io"
```
MancunianSam, line 2: Can you name it `"${local.environment}-dr2-ingest-check-preservica-for-existing-io"` please.

MancunianSam (commented):
Looks good. Just one comment. Can you add the lambda arn to the `deploy_roles` file as well please.

MancunianSam (approved):

PR #97, opened 2024-04-23
DR2-1591 Add new queue to topic subscription.

MancunianSam:
For the prod queue linked to the TRE topic, I've added a num
subscription to the new queue. Before I deploy this, I'll disable the
trigger for the court document lambda. Then if any messages come through
while this subscription is being added, we won't lose them.

If I just change the subscription, there'll be a tiny window where
there'll be no subscription and sod's law says we'll get a message then
so this is safer.

For the anonymiser queue, we don't care if we lose a message so this is
just being replaced.


techncl (approved):
Looks good!

PR #98, opened 2024-04-23
Fixes for things I missed the first time

MancunianSam:
The new anonymiser lambda was writing to the wrong bucket and sending a
message with the old bucket.

The KMS key didn't have the new anonymiser lambda in it's list of
allowed decrypt roles.

The new parsed document handler was starting the old step function.


techncl (dismissed):
👍

techncl (approved):

PR #100, opened 2024-04-24
DR2-1523 Add notifications SNS topic.

MancunianSam:
The topic policy allows same account subscribe and publish. We can
change this later.

I've added the new KMS key which allows the admin user to decrypt in
intg and no-one to decrypt in staging and prod.


techncl (commented):
Looks good. I was thinking, maybe `notifications.tf` could be renamed to `sns_notifications.tf` or `sns_topic_notifications.tf` so they are not confused with any other notifications e.g. slack notifications?

MancunianSam:
It's a fair point. I've renamed it to `sns_notifications.tf`

techncl (approved):
👍

PR #101, opened 2024-04-24
DR2-1563 Add new GSI on ingested_PS

MancunianSam:
Sadly we can't have a boolean index so this is a string and we'll need
to rely on the DynamoFormatters to make sure this doesn't get into a
mess.

The rest of the changes are renaming and reformatting as we've got more
than one index now.


techncl (approved):
👍

PR #104, opened 2024-05-01
Add lock table env to asset reconciler

techncl:


ingest_asset_reconciler.tf
```diff
@@ -23,6 +23,7 @@ module "dr2_ingest_asset_reconciler_lambda" {
     PRESERVICA_API_URL     = data.aws_ssm_parameter.preservica_url.value
     DYNAMO_TABLE_NAME      = local.files_dynamo_table_name
```
kurtismash, line 24: Is it worth renaming this to `DYNAMO_FILES_TABLE_NAME`? How painful is doing that?

techncl: Hmm, I guess, in order to avoid disruption, we could add another env variable underneath `DYNAMO_TABLE_NAME`, like this `DYNAMO_FILES_TABLE_NAME = local.files_dynamo_table_name`, deploy it, change it in the Lambda, deploy that and then remove the `DYNAMO_TABLE_NAME` line; do you think that would work?

techncl: `DYNAMO_TABLE_NAME` is also in other lambda, should we change it there too?

kurtismash (approved):

PR #105, opened 2024-05-01
Add SNS publish step to ingest step function

techncl:


templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -462,13 +462,24 @@
...
+            "Parameters": {
+              "Message": {
+                "reconciliationUpdate": "Asset was reconciled",
+                "assetId.$": "$.AssetInfo.assetId",
+                "messageId": "$.messageId",
+                "parentMessageId": "$.parentMessageId",
+                "executionId": "$.executionId"
+              },
+              "TopicArn": "arn:aws:sns:eu-west-2:${account_id}:${ingest_dr2_notifications_name}"
+            },
```
kurtismash, line 482: How difficult would it be to wrap this in a MAP state and send 10 at a time using the [PublishBatch](https://docs.aws.amazon.com/sns/latest/api/API_PublishBatch.html) endpoint?

templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -462,13 +462,24 @@
...
+          "Publish reconciliation update to dr2-notifications": {
+            "Type": "Task",
+            "Resource": "arn:aws:states:::sns:publish",
+            "Parameters": {
+              "Message": {
+                "reconciliationUpdate": "Asset was reconciled",
+                "assetId.$": "$.AssetInfo.assetId",
+                "messageId": "$.messageId",
+                "parentMessageId": "$.parentMessageId",
+                "executionId": "$.executionId"
```
kurtismash, line 479: Can you pop these into a nested `properties` object? We've got more to do on the schema but this is a good start.

kurtismash: Also missing the trailing `.$` on the keys.

templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -462,13 +462,24 @@
...
-          "Do nothing": {
-            "Type": "Pass",
+          "Publish reconciliation update to dr2-notifications": {
+            "Type": "Task",
+            "Resource": "arn:aws:states:::sns:publish",
+            "Parameters": {
+              "Message": {
+                "reconciliationUpdate": "Asset was reconciled",
+                "assetId.$": "$.AssetInfo.assetId",
+                "messageId": "$.messageId",
```
kurtismash, line 477: Where does this UUID come from? Is there any benefit to us using `"field.$": "States.UUID()"` instead ([link](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-intrinsic-functions.html#asl-intrsc-func-uuid-generate))?

common.tf
```diff
@@ -199,6 +200,7 @@ module "dr2_ingest_step_function" {
     ingest_start_workflow_lambda_name             = local.ingest_start_workflow_lambda_name
     ingest_workflow_monitor_lambda_name           = local.ingest_workflow_monitor_lambda_name
     ingest_asset_reconciler_lambda_name           = local.ingest_asset_reconciler_lambda_name
+    ingest_dr2_notifications_name                 = local.notifications_topic_name
```
kurtismash, line 203: I don't think we need the `ingest_dr2` prefix on the var name.

common.tf
```diff
@@ -275,6 +277,7 @@ module "dr2_ingest_step_function_policy" {
...
```
kurtismash, line 280: I don't think we need the `ingest_dr2` prefix on the var name.

templates/sfn/ingest_sfn_definition.json.tpl
techncl: Because you wrote in the "DA-ADR-13 Notification Message Tracing" doc that "All new notification message schemas should include `messageId`, `parentMessageId`, and `executionId`...". I thought these 3 would be in the `message` and have to be extracted

kurtismash: `messageId` in the message we receive would become `parentMessageId` in the message we send. `executionId` in our case will be the Step Function execution name.

templates/sfn/ingest_sfn_definition.json.tpl
techncl: But isn't this already in a MAP? What would we be mapping over in this case?

kurtismash: Ahh, that makes it difficult. Let's drop this requirement until we need it then. It's $0.50/million requests which would be a nice cost saving when we do migrations.

Follow up but happy for this to be a backlog ticket - if the reconciler received a list of asset ids, would it cope and output a value per asset? I guess the other option is to map(batch=1) within a map(batch=10).

techncl: It can't handle a list of asset Ids in its current state but I guess it could, although, it would be a little messy since you'd have to wrap the current Lambda code in a map. Might be worth creating a ticket so that we could discuss it.

techncl:
I'm just going to wait for the [reconcile Lambda PR](https://github.com/nationalarchives/dr2-ingest/pull/78/) to go in so that I can test the SNS/terraform changes before I make them

templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -462,19 +462,39 @@
...
+                "properties": {
+                  "messageId": "$.reconciliationSnsMessage.properties.messageId",
+                  "parentMessageId": "$.reconciliationSnsMessage.properties.parentMessageId",
+                  "timestamp": "$.reconciliationSnsMessage.properties.timestamp",
+                  "type": "preserve.digital.asset.ingest.update"
+                },
+                "parameters": {
+                  "assetId": "$.reconciliationSnsMessage.parameters.assetId",
+                  "status": "Asset has been ingested to the Preservation system."
+                }
```
kurtismash, line 484: Just to confirm the mappings upstream of this...
- `$.reconciliationSnsMessage.properties.messageId` is a new UUID?
- `$.reconciliationSnsMessage.properties.parentMessageId` is the `messageId` of the message in the lock table?
- `$.reconciliationSnsMessage.properties.timestamp` is an ISO timestamp?
- `$.reconciliationSnsMessage.parameters.assetId` is the `name` of the Asset item in dr2-ingest-files?

kurtismash: Yes please, we're in a bit of an odd situation here. Currently the `id` in our table is generated during our preingest lambda so only we recognise it. We have a ticket in the backlog to use the TDR UUID as the `id` but can't do that until we've changed our table's primary key. So for now we need to use the `name`.

kurtismash (approved):

PR #106, opened 2024-05-04
Confirm Lock table has no items with speficied batchId

techncl:


common.tf
```diff
@@ -199,6 +199,8 @@ module "dr2_ingest_step_function" {
     ingest_start_workflow_lambda_name             = local.ingest_start_workflow_lambda_name
     ingest_workflow_monitor_lambda_name           = local.ingest_workflow_monitor_lambda_name
     ingest_asset_reconciler_lambda_name           = local.ingest_asset_reconciler_lambda_name
+    ingest_lock_table_name                        = local.ingest_lock_dynamo_table_name
+    ingest_lock_table_gsi_name                    = local.ingest_lock_table_global_secondary_index_name
```
kurtismash, line 203: Each ddb table can have up to 20 GSIs, to be future flexible is there a better way to name this var?

templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -495,7 +495,42 @@
...
+        }
+      },
+      "Resource": "arn:aws:states:::aws-sdk:dynamodb:query",
+      "Next": "Check if number of items is 0"
+    },
+    "Check if number of items is 0": {
+      "Type": "Choice",
+      "Choices": [
+        {
+          "Variable": "$.ScannedCount",
```
kurtismash, line 519: Although `Count` and `ScannedCount` [will be the same](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#API_Query_ResponseElements:~:text=If%20you%20did%20not%20use%20a%20filter%20in%20the%20request%2C%20then%20ScannedCount%20is%20the%20same%20as%20Count.) as we're not applying a filter, can we use `Count` here to future-proof a filter being added.

PR #107, opened 2024-05-14
DR2-1639 Update Github actions role trust policy.

MancunianSam:
These should be limited only to the environments and repositories that
they need


techncl (approved):
Looks good!

PR #124, opened 2024-06-11
DR2-1715 Create lambda with new name.

MancunianSam:
In an attempt not to make breaking changes, I'm creating the lambda with
the new name first.
Then I'll change dr2-ingest to deploy to this new lambda.
Then I'll switch over the step functions to the new one and delete the
old one.

### DIFF
--- a/ingest_check_preservica_for_existing_io.tf
+++ b/ingest_check_preservica_for_existing_io.tf
@@ -1,6 +1,6 @@
 locals {
   ingest_check_preservica_for_existing_io_lambda_name_old = "${local.environment}-dr2-ingest-check-preservica-for-existing-io"
-  ingest_check_preservica_for_existing_io_lambda_name = "${local.environment}-dr2-ingest-find-existing-asset"
+  ingest_check_preservica_for_existing_io_lambda_name     = "${local.environment}-dr2-ingest-find-existing-asset"
 }
 module "ingest_check_preservica_for_existing_io_lambda" {
   source          = "git::https://github.com/nationalarchives/da-terraform-modules//lambda"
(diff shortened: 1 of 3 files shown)


ingest_check_preservica_for_existing_io.tf
```diff
@@ -1,7 +1,41 @@
 locals {
-  ingest_check_preservica_for_existing_io_lambda_name = "${local.environment}-dr2-ingest-check-preservica-for-existing-io"
+  ingest_check_preservica_for_existing_io_lambda_name_old = "${local.environment}-dr2-ingest-check-preservica-for-existing-io"
+  ingest_check_preservica_for_existing_io_lambda_name     = "${local.environment}-dr2-ingest-find-existing-asset"
```
techncl, line 3: Does this need to have the old name?

techncl (commented):
Looks good! Just a minor question.

ingest_check_preservica_for_existing_io.tf
MancunianSam: It does at the moment because the old lambda is still used in the step function. Once the code is deployed to the new lambda, we can switch the step function over to using the new lambda and delete the old name.

techncl: Sorry, I meant the variable name `ingest_check_preservica_for_existing_io_lambda_name ` not the lambda name

MancunianSam: Ah I see. Yeah, good point, I've renamed it. I'll rename the file when I delete the old lambda.

techncl (approved):
👍

PR #126, opened 2024-06-19
DR2-1259 Add secrets rotation lambda

MancunianSam:


techncl (commented):
Looks good. Just two questions:

1. Should this be added to the `dashboard_lambdas` list?
2. The `CreatedBy` tag has `dp-terraform-environments` as a value, and other modules seem to have it but we also have a (unused) `creator` variable in the local variables section in the `common.tf` file, with the value of `dr2-terraform-environments`; which one is correct?

MancunianSam:
1. Good point, it should I think. I've added it in.
2. The creator local variable is right. That's used in the `root_provider.tf` file to add a default `CreatedBy` tag which should add that to all resources. Adding a `CreatedBy` tag on anything else is pointless so I've taken them all out. Good spot, cheers.

techncl (approved):
👍

PR #127, opened 2024-06-19
Small terraform fixes

MancunianSam:
* The dev kms key was using the old variable and so didn't have
  permissions for s3/sns etc.
* The newest da-terraform-modules adds the SSL only block for buckets so
  we don't need it here.
* The preservica config SNS topic had the wrong name and a mismatch
  between the resource in its policy and its name.


techncl (approved):
👍

PR #130, opened 2024-06-26
DR2-1759 Remove topic encryption

MancunianSam:


kurtismash (approved):

kurtismash:
I'm not sure how TF handles removing a KMS key as there's a wait period after marking for deletion. We may need to check the key's gone in 30 days.

PR #132, opened 2024-06-27
DR2-1730 DR2-1750 Rename table, change primary key, update SFN

MancunianSam:
None of these will work seperately so I'm lumping them into the same change


kurtismash (approved):

common.tf
```diff
@@ -288,6 +288,7 @@ module "dr2_ingest_step_function_policy" {
 module "files_table" {
```
techncl, line 288: Should we change this to `ingest_files_table` or `files_ingest_table`?

techncl, line 288: Should we change this to `ingest_files_table` or `files_ingest_table`?

PR #134, opened 2024-07-02
Dr2 1662 create dr aws infrastructure

sparkhi:


custodial_copy_notifications.tf
```diff
@@ -0,0 +1,54 @@
+locals {
+  custodial_copy_topic_name  = "${local.environment}-dr2-cc-notifications"
+  custodial_copy_topic_arn   = "arn:aws:sns:eu-west-2:${data.aws_caller_identity.current.account_id}:${local.custodial_copy_topic_name}"
+  custodial_copy_queue_name  = "${local.environment}-dr2-cc-notifications"
+  custodial_copy_lambda_name = "${local.environment}-dr2-cc-notifications"
```
kurtismash, line 5: These two should be `${local.environment}-dr2-ingest-cc-notifications`, and there variable names as well. The SNS topic is part of our CC worker and we're using it to fan-out the events to other parts of our application. The SQS queue and Lambda we're implementing at the moment are part of our ingest pipeline. We may want to implement more subscribers in future.

```mermaid
flowchart LR;
    subgraph ingest
        sqs[dr2-ingest-cc-notifications Queue] --> l[dr2-ingest-cc-notifications Lambda]--> ddb[dr2-ingest-files]
    end
    subgraph custodial-copy
        sns[dr2-cc-notifications Topic]
    end
    sns --> sqs
```

custodial_copy_notifications.tf
```diff
@@ -0,0 +1,54 @@
...
+module "dr2_custodial_copy_queue" {
+  source     = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
+  queue_name = local.custodial_copy_queue_name
+  sqs_policy = templatefile("./templates/sqs/sns_send_message_policy.json.tpl", {
+    account_id = var.account_number,
+    queue_name = local.custodial_copy_queue_name
+    topic_arn  = local.custodial_copy_topic_arn
+  })
+  encryption_type = "sse"
+}
```
kurtismash, line 27: The DLQ for this will need adding to the alerting rule. https://github.com/nationalarchives/dr2-terraform-environments/blob/e5d8ecc60592a27374eee8223f871d97f512c76b/common.tf#L341-L363.

custodial_copy_notifications.tf
```diff
@@ -0,0 +1,54 @@
...
+
+resource "aws_sns_topic_subscription" "dr2_custodial_copy_queue_subscription" {
```
kurtismash, line 29: Please can `raw_message_delivery` be enabled on this, it avoids double encapsulation when we receive the messages via SQS. I forgot to put this in the ticket.

Docs: https://docs.aws.amazon.com/sns/latest/dg/sns-large-payload-raw-message-delivery.html#raw-message-examples

templates/iam_policy/custodial_copy_lambda_policy.json.tpl
```diff
@@ -0,0 +1,37 @@
+{
+  "Statement": [
+    {
+      "Action": [
+        "dynamodb:UpdateItem",
+        "dynamodb:BatchGetItem"
```
kurtismash, line 6: I think we're going to need `dynamodb:query` here.

kurtismash (commented):
Looks good @sparkhi, only a few of comments.

custodial_copy_notifications.tf
```diff
@@ -0,0 +1,54 @@
...
+  topic_arn = local.custodial_copy_topic_arn
+  protocol  = "sqs"
+  endpoint  = module.dr2_custodial_copy_queue.sqs_arn
+}
+
+module "dr2_custodial_copy_lambda" {
```
techncl, line 35: Maybe this should be `dr2_custodial_copy_notifications_lambda`?

custodial_copy_notifications.tf
```diff
@@ -0,0 +1,54 @@
...
+  source        = "git::https://github.com/nationalarchives/da-terraform-modules//lambda"
+  function_name = local.custodial_copy_lambda_name
+  handler       = "uk.gov.nationalarchives.custodialcopy.Lambda::handleRequest"
```
MancunianSam, line 38: This will be a Python lambda so you'll need to update the `handler`, `timeout_seconds`, `memory_size` and `runtime` variables.
There's an example here https://github.com/nationalarchives/dr2-terraform-environments/blob/main/ip_lock_checker_lambda.tf#L16

custodial_copy_notifications.tf
MancunianSam: You don't need the `aws_sns_topic_subscription` resource here at all. This can be deleted and you can add a block to the topic module.
```hcl
module "dr2_custodial_copy_topic" {
  source = "git::https://github.com/nationalarchives/da-terraform-modules//sns"
  ...
  sqs_subscriptions = {
    custodial_copy_queue = module.dr2_custodial_copy_queue.sqs_arn
  }
}
```
Raw message delivery is set by default for sqs subscriptions as well.

sparkhi: That's correct, that's how it was in the entity even generator which I used as template. However, I felt the SNS should not be the one initiating subscriptions hence I kept it separate. I'm not fussed about it, if you feel it's better to keep it with SNS, I can change it like that.

custodial_copy_notifications.tf
sparkhi: Changed to `dr2_custodial_copy_ingest_lambda` - similar to the name

custodial_copy_notifications.tf
sparkhi: Ignore the comment above, changed it to be consistent with previous

kurtismash:
What's our precedent for making a Terraform file? Should each Lambda be its own file with it's sources or are we happy putting the SNS, SQS, and Lambda in a file for the SNS?

sparkhi:
> What's our precedent for making a Terraform file? Should each Lambda be its own file with it's sources or are we happy putting the SNS, SQS, and Lambda in a file for the SNS?



> What's our precedent for making a Terraform file? Should each Lambda be its own file with it's sources or are we happy putting the SNS, SQS, and Lambda in a file for the SNS?

I looked at the "entity event generator" file as a reference and that has lambda, sns and sqs in the same file

MancunianSam:
I think it's fine as is. Each file is a "component" including all the bits around it that make it work so I think having the SNS topic that feeds the SQS -> Lambda in the same file is OK.

MancunianSam (approved):

PR #136, opened 2024-07-04
DR2-1769 Create SQS queue

MancunianSam:
* Add the DLQ to the alarm.
* Add the new repos to the DR user
* Subscribe the new queue to the DR notifications topic


kurtismash (approved):
Approving but a couple of open questions
- DR or CC, we're using CC for the notifications but the queue being created is DR.
- Does the same IAM user run both on the bare metal and the docker container?

MancunianSam:
1. It's going in the existing disaster_notifications file so I called it DR. I'll rename everything at some point but not now.
2. Yes

kurtismash (approved):
2. We should fix that.

PR #137, opened 2024-07-08
DR2-1772 Secrets rotation and notifications.

MancunianSam:
We get one for RotationFailed and RotationSucceeded.

Because of the way the transformers work I've not included the error
message as it makes the terraform messy. This is find because Slack is
not a debugging tool.


techncl (approved):
Looks good!

PR #138, opened 2024-07-10
DR2-1755 Log external notifications

MancunianSam:
Anything going to the notifications topic will go
SNS -> SQS -> Eventbridge pipe -> Cloudwatch.

There is an input transformer configured so we don't get all the output
from the SQS message.

Given an SNS message:

```json
{
  "id": 1,
  "message": "Another test message"
}

```

You will end up with this in Cloudwatch:
```json
{
    "body": {
        "id": 1,
        "message": "Another test message"
    },
    "timestamp": "1720607411035"
}
```

### DIFF
--- a/external_notifications_log.tf
+++ b/external_notifications_log.tf
@@ -1,10 +1,8 @@
 locals {
-  external_notifications_queue_name     = "${local.environment}-external-notifications"
-  external_notifications_pipe_name      = "${local.environment}-external-notifications"
-  external_notifications_log_group_name = "/${local.environment}-external-notifications"
+  external_notifications_name = "${local.environment}-external-notifications"
 }
 resource "aws_cloudwatch_log_group" "external_notification_log_group" {
-  name = local.external_notifications_log_group_name
+  name = "/${local.external_notifications_name}"
 }
 
 module "dr2_external_notifications_pipes_role" {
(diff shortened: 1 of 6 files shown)


external_notifications_log.tf
```diff
@@ -0,0 +1,53 @@
+locals {
+  external_notifications_queue_name     = "${local.environment}-external-notifications"
```
techncl, line 2: Since these have essentially the same value, could there be one main variable that these 3 make reference to?

techncl (commented):
Looks good! Just one question. As well as

I was wondering, is it worth adding (for filtering purposes):

1. what sent the notification
2. the topic arn

external_notifications_log.tf
MancunianSam: Yeah ok. I've changed it to `external_notifications_name`

MancunianSam:
I don't think we're fussed about what sent the notification and it's possible we'll have more than one thing sending to the topic so I'll leave that out. The topicArn makes sense though, I've added that in.

techncl (approved):
👍

external_notifications_log.tf
```diff
@@ -0,0 +1,53 @@
+locals {
+  external_notifications_name = "${local.environment}-external-notifications"
```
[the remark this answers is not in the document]
MancunianSam: Yes, I suppose there is but does that matter as long as we know what the log group is called?

PR #154, opened 2024-07-31
Run cc-notifications

MancunianSam:
This adds it to the deploy policy and the role trust policy and sets up
the sqs queue mapping.


techncl (approved):
👍

PR #156, opened 2024-08-05
Dr2 1804 rename ioId and batchId in lock table

techncl:


MancunianSam (approved):
Looks good. I think you need to update the Dynamo formatters https://github.com/nationalarchives/dr2-ingest/blob/main/dynamo-formatters/src/main/scala/uk/gov/nationalarchives/dynamoformatters/DynamoFormatters.scala#L251 before this can be deployed.

MancunianSam (approved):

PR #157, opened 2024-08-05
Remove notifications publish.

MancunianSam:
This step is now being done by the files change handler. Or it should
be, I'm not convinced it's working but it's harder to debug with this
sending as well and we won't need this if the files change handler
works properly so it can be removed.


techncl (approved):
👍

PR #158, opened 2024-08-06
Remove secrets manager success notification.

MancunianSam:
These are annoying and not necessary. The failed ones will still alert.


techncl (approved):
👍

PR #160, opened 2024-08-06
DR2-1839 Terraform for copy tdr files

MancunianSam:
This includes:

* The queue
* The lambda with a subscription to the queue
* A subscription to  TDRs SNS with a filter. Standard and judgment
  transfers get sent here. This filters out the judgments.
* Access to the KMS key. The lambda writes to the raw cache bucket which
  is KMS encrypted.


copy_files_from_tdr.tf
```diff
@@ -0,0 +1,56 @@
...
+  encryption_type          = local.sse_encryption
+}
+
+resource "aws_sns_topic_subscription" "dr2_copy_files_from_tdr_subscription" {
+  endpoint             = module.dr2_copy_files_from_tdr_sqs.sqs_arn
+  protocol             = "sqs"
+  topic_arn            = local.tdr_external_notifications_topic
+  raw_message_delivery = true
+  filter_policy_scope  = "MessageBody"
+  filter_policy        = templatefile("${path.module}/templates/sns/tdr_filter_policy.json.tpl", { environment = local.environment })
```
techncl, line 55: On line 18, there is a `tdr_bucket_name`, could that be saved to a local variable and then passed into this template file instead of the environment (and then change the template file to accept the bucket name)?

copy_files_from_tdr.tf
```diff
@@ -0,0 +1,56 @@
...
+module "dr2_copy_files_from_tdr_sqs" {
+  source     = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
+  queue_name = local.copy_files_from_tdr_name
+  sqs_policy = templatefile("./templates/sqs/sns_send_message_policy.json.tpl", {
+    account_id = var.account_number,
+    queue_name = local.copy_files_from_tdr_name
+    topic_arn  = "arn:aws:sns:eu-west-2:${module.tdr_config.account_numbers[local.environment]}:tdr-external-notifications-${local.environment}"
+  })
+  redrive_maximum_receives = 5
+  visibility_timeout       = 180
```
techncl, line 45: These 2 variables are used a few times in the modules of this repo; should we add them to local variables?

copy_files_from_tdr.tf
```diff
@@ -0,0 +1,56 @@
...
+}
+
+
+module "dr2_copy_files_from_tdr_sqs" {
+  source     = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
+  queue_name = local.copy_files_from_tdr_name
+  sqs_policy = templatefile("./templates/sqs/sns_send_message_policy.json.tpl", {
+    account_id = var.account_number,
+    queue_name = local.copy_files_from_tdr_name
+    topic_arn  = "arn:aws:sns:eu-west-2:${module.tdr_config.account_numbers[local.environment]}:tdr-external-notifications-${local.environment}"
```
techncl, line 42: Could you use `local.tdr_external_notifications_topic` here?

techncl (commented):
Looks good! Just a few comments

copy_files_from_tdr.tf
MancunianSam: Good idea. I've renamed the template file `bucket_filter_policy.json.tpl` as we could theoretically use if for something else now.

copy_files_from_tdr.tf
MancunianSam: Yeah ok. I've made two new variables and replaced them where I could find them.

copy_files_from_tdr.tf
MancunianSam: Oh yeah. Knew I missed one :smile:

techncl (approved):
👍

PR #161, opened 2024-08-09
Add permissions for TDR KMS.

MancunianSam:
When you're doing cross account KMS, you need permissions on both ends.
I've updated TDR so it needs doing in the lambda policy


techncl (approved):
👍

PR #162, opened 2024-08-09
DR2-1847 Add reindexer image to user policy.

MancunianSam:
This will allow the CC user to pull the new reindexer docker image.


techncl (approved):
👍

PR #165, opened 2024-08-28
Link batch variables to env vars.

MancunianSam:
The environment variables of the lambda should be linked to the batch
configuration so that the lambda code knows when to create a new batch.

### DIFF
--- a/tdr_preingest.tf
+++ b/tdr_preingest.tf
@@ -1,13 +1,13 @@
 locals {
-  tdr_preingest_name              = "${local.environment}-dr2-preingest-tdr"
-  tdr_aggregator_name             = "${local.tdr_preingest_name}-aggregator"
-  tdr_aggregator_queue_arn        = "arn:aws:sqs:eu-west-2:${data.aws_caller_identity.current.account_id}:${local.tdr_aggregator_name}"
-  tdr_aggregator_batching_window  = 300
-  tdr_aggregator_batch_size       = 10000
-  tdr_aggregator_queue_url        = "https://sqs.eu-west-2.amazonaws.com/${data.aws_caller_identity.current.account_id}/${local.tdr_aggregator_name}"
-  tdr_package_builder_lambda_name = "${local.tdr_preingest_name}-package-builder"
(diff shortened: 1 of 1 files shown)


tdr_preingest.tf
```diff
@@ -2,6 +2,8 @@ locals {
   tdr_preingest_name              = "${local.environment}-dr2-preingest-tdr"
   tdr_aggregator_name             = "${local.tdr_preingest_name}-aggregator"
   tdr_aggregator_queue_arn        = "arn:aws:sqs:eu-west-2:${data.aws_caller_identity.current.account_id}:${local.tdr_aggregator_name}"
+  tdr_aggregator_batching_window  = 300
+  tdr_aggregator_batch_size       = 10000
```
techncl, line 6: Is this Bytes or Kilobytes,  Maybe "bytes" or "kb" could be appended to the name?

tdr_preingest.tf
```diff
@@ -2,6 +2,8 @@ locals {
   tdr_preingest_name              = "${local.environment}-dr2-preingest-tdr"
   tdr_aggregator_name             = "${local.tdr_preingest_name}-aggregator"
   tdr_aggregator_queue_arn        = "arn:aws:sqs:eu-west-2:${data.aws_caller_identity.current.account_id}:${local.tdr_aggregator_name}"
+  tdr_aggregator_batching_window  = 300
```
techncl, line 5: Maybe "in_secs" could be appended to this name?

techncl (commented):
Looks good! Just two minor comments

tdr_preingest.tf
MancunianSam: It's neither, it's the number of items in the queue. AWS call it batch size to I'll keep it so they match.

tdr_preingest.tf
MancunianSam: I'll go with `tdr_aggregator_batching_window_seconds` to match some of the other ones.

techncl (approved):
👍

PR #166, opened 2024-08-29
DR2-1859 Remove kms encryption from the lock table

MancunianSam:
Setting this to false changes the encryption to AWS managed.

### DIFF
--- a/common.tf
+++ b/common.tf
@@ -331,8 +331,7 @@ module "ingest_lock_table" {
   source                         = "git::https://github.com/nationalarchives/da-terraform-modules//dynamo"
   hash_key                       = { name = local.ingest_lock_table_hash_key, type = "S" }
   table_name                     = local.ingest_lock_dynamo_table_name
-  server_side_encryption_enabled = true
-  kms_key_arn                    = module.dr2_kms_key.kms_key_arn
+  server_side_encryption_enabled = false
   additional_attributes = [
     { name = "groupId", type = "S" }
   ]


common.tf
```diff
@@ -331,8 +331,7 @@ module "ingest_lock_table" {
   source                         = "git::https://github.com/nationalarchives/da-terraform-modules//dynamo"
   hash_key                       = { name = local.ingest_lock_table_hash_key, type = "S" }
   table_name                     = local.ingest_lock_dynamo_table_name
-  server_side_encryption_enabled = true
-  kms_key_arn                    = module.dr2_kms_key.kms_key_arn
+  server_side_encryption_enabled = false
```
techncl, line 334: I got confused by the message you wrote, you wrote "Setting this to false changes the encryption to AWS managed." but it says [here](https://github.com/terraform-aws-modules/terraform-aws-dynamodb-table?tab=readme-ov-file#input_server_side_encryption_enabled) that this setting is "Whether or not to enable encryption at rest using an AWS managed KMS customer master key (CMK)" so wouldn't setting to `false` (default) do the opposite of what you said?

techncl (commented):
Looks good! Just one comment

common.tf
MancunianSam: That link misses some bits. This https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dynamodb_table#server_side_encryption says that setting it to false uses the AWS managed encryption.

techncl (approved):
👍

PR #169, opened 2024-09-05
Rename preingest packages

MancunianSam:
This matches the lambda and repo directory names

Also, add missing PutItem permission to the aggregator lambda role.


techncl (approved):
👍

PR #170, opened 2024-09-16
Create Validation Lambda

techncl:


ingest_validate_generic_ingest_inputs.tf
```diff
@@ -0,0 +1,22 @@
+locals {
+  ingest_validate_generic_ingest_inputs_lambda_name = "${local.environment}-dr2-ingest-validate-generic-ingest-inputs"
```
kurtismash, line 2: I think we'll get away with calling this `<env>-dr2-ingest-validator` as it's scoped to the ingest part of the app through `dr2-ingest`.

templates/iam_policy/ingest_validate_generic_ingest_inputs_policy.json.tpl
```diff
@@ -0,0 +1,31 @@
+{
+  "Statement": [
+    {
+      "Action": [
+        "s3:GetObject",
+        "s3:HeadObject",
```
kurtismash, line 6: There isn't an action for `s3:HeadObject`, it uses `s3:GetObject` - https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html.

AWS API endpoints != IAM actions 😵

templates/s3/lambda_access_bucket_policy.json.tpl
```diff
@@ -8,6 +8,7 @@
       },
       "Action": [
         "s3:GetObject",
+        "s3:HeadObject",
```
kurtismash, line 11: See other comment.

templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -1,7 +1,34 @@
 {
   "Comment": "${step_function_name}: A State machine to ingest DR2 BagIt-like packages into Preservica.",
-  "StartAt": "Map metadata",
+  "StartAt": "Validate the metadata.json file",
```
kurtismash, line 3: Can we call this `Validate input` instead as it should be doing both the `metadataPackage` JSON and the state machine input.

ingest_validate_generic_ingest_inputs.tf
```diff
@@ -0,0 +1,22 @@
...
+      account_id  = var.account_number
+      lambda_name = local.ingest_validate_generic_ingest_inputs_lambda_name
+    })
+  }
+  memory_size = local.java_lambda_memory_size
+  runtime     = local.java_runtime
+  tags = {
+    Name = local.ingest_validate_generic_ingest_inputs_lambda_name
+  }
+}
```
kurtismash, line 22: Does this need any env vars to know where to put files that fail validation?

kurtismash (commented):
Looks good, just some comments on naming and a fun quirk of AWS. 👍

ingest_validate_generic_ingest_inputs.tf
techncl: No, as it will upload a JSON (with just the entries that failed) to the same S3 bucket that the original metadata.json is, with a different name.

ingest_validate_generic_ingest_inputs.tf
techncl: Wait, I think the Lambda has to be the same name as the repo

MancunianSam (approved):

PR #171, opened 2024-09-17
DR2-1776 Use deleteItem to remove lock table row.

MancunianSam:
Also, use assetId. It's the same as assetName in the input but this
makes more sense.


techncl (approved):
👍

PR #173, opened 2024-09-18
Retry ingests if Assets are still in lock table

techncl:


templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -559,12 +559,38 @@
...
+    },
+    "Retry pre-ingest step function": {
+      "Type": "Task",
+      "Resource": "arn:aws:states:::states:startExecution",
+      "Parameters": {
+        "StateMachineArn.$": "$$.Execution.Input.retrySfnArn",
+        "Name.$": "$$.Execution.Name",
+        "Input": {
+          "groupId.$": "$$.Execution.Input.groupId",
+          "batchId.$": "States.Format('{}_{}', $$.Execution.Input.groupId, States.MathAdd($$.Execution.Input.retryCount, 1))",
```
kurtismash, line 587: The execution name will need to match the `batchId`. The preingest step function [starts this step function](https://github.com/nationalarchives/dr2-terraform-environments/blob/28fcb69dff6cc102b517a7b78047ee6beb446329/templates/sfn/preingest_tdr_sfn_definition.json.tpl#L45C9-L45C38) using `"Name.$": "$$.Execution.Name"`, so we'll see an error as the execution already exists.

MancunianSam (approved):

PR #175, opened 2024-09-23
Ignore the enabled status for the copy files queue.

MancunianSam:
This lets us stop TDR transfers coming in. The messages will just sit in
the queue.


techncl (approved):
👍

PR #176, opened 2024-09-23
Create 'dr2-ingest-state' bucket

techncl:


common.tf
```diff
@@ -240,6 +242,19 @@ module "dr2_ingest_step_function" {
...
+  source      = "git::https://github.com/nationalarchives/da-terraform-modules//s3"
+  bucket_name = local.ingest_state_bucket_name
+  bucket_policy = templatefile("./templates/s3/lambda_access_bucket_policy.json.tpl", {
+    lambda_role_arns = jsonencode([
+      module.dr2_ingest_asset_opex_creator_lambda.lambda_role_arn,
+      module.dr2_ingest_folder_opex_creator_lambda.lambda_role_arn,
+      module.dr2_ingest_asset_reconciler_lambda.lambda_role_arn
+    ]),
+    bucket_name = local.ingest_state_bucket_name
+  })
```
kurtismash, line 255: Only the Mapper Lambda and Step Function itself will need access to the bucket (read/write and read respectively).

Whilst not required as the bucket is in the same account as the resources, the bucket should also be added to the permission policy of the roles used by the Lambda and Step Function. Normally, resource policies are reserved for managing service access to a resource or cross account.

common.tf
```diff
@@ -240,6 +242,19 @@ module "dr2_ingest_step_function" {
...
+  kms_key_arn = module.dr2_kms_key.kms_key_arn
```
kurtismash, line 256: It's likely we'll want to see/edit the data in these S3 objects. As the objects will only contain arrays of UUIDs we should be able to use the developer key here.

NB. We'll use the dev key instead of SSE-S3 just in case we put some error info here in future.

techncl: Ah, I think I added it because you left this comment on the ticket "Some limited error information may each this bucket, use the dev KMS key?"

kurtismash: We have 2 keys in our project. We encrypt most with [`dr2_kms_key`](https://github.com/nationalarchives/dr2-terraform-environments/blob/d77177701409b11507b00d952a9fcae6a6bf215f/common.tf#L128) as this allows only the system principals to decrypt, but we also have [`dr2_developer_key`](https://github.com/nationalarchives/dr2-terraform-environments/blob/d77177701409b11507b00d952a9fcae6a6bf215f/common.tf#L161) which still encrypts with a key we manage permissions on but also allows developers to decrypt.

Using the developer key will allow us to see error data and triage without granting the ability to see all data within the system.

kurtismash (approved):

PR #67, opened 2024-01-18
DR2-1414 Update the parsed document queue.

MancunianSam:
This was originally allowing the same account to use the queue. It has
been updated to allow the prod TRE SNS topic to use it.

PR #177, opened 2024-09-24
DR2-1889 FIFO queue and creator lambda

MancunianSam:
This creates the lambda which will add the IO id as the message group id
and creates the queue which triggers it.

I've updated the entity event topic to send to the new queue and updated
the existing custodial copy queue to be a fifo queue.

I've added the new sqs queue to the alarms and added the lambda to the
deploy policy.

This needs
https://github.com/nationalarchives/da-terraform-modules/pull/67 before
this will work.

### DIFF
--- a/common.tf
+++ b/common.tf
@@ -43,7 +43,8 @@ locals {
     local.rotate_preservation_system_password_name,
     local.ingest_start_workflow_lambda_name,
     local.ingest_upsert_archive_folders_lambda_name,
-    local.ingest_workflow_monitor_lambda_name
+    local.ingest_workflow_monitor_lambda_name,
+    local.ingest_queue_creator_name
   ]
 }
 resource "random_password" "preservica_password" {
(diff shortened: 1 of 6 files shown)


custodial_copy.tf
```diff
@@ -22,7 +22,7 @@ resource "aws_iam_group" "custodial_copy_group" {
 
 resource "aws_iam_group_policy" "custodial_copy_group_policy" {
   group = aws_iam_group.custodial_copy_group.name
-  name  = "${local.environment}-dr2-custodial-copy-policy"
+  name  = "${local.environment}-dr2-c.ustodial-copy-policy"
```
techncl, line 25: This seems like a mistake

common.tf
```diff
@@ -43,7 +43,8 @@ locals {
     local.rotate_preservation_system_password_name,
     local.ingest_start_workflow_lambda_name,
     local.ingest_upsert_archive_folders_lambda_name,
-    local.ingest_workflow_monitor_lambda_name
+    local.ingest_workflow_monitor_lambda_name,
+    local.ingest_queue_creator_name
```
techncl, line 47: Should this have "lambda" in its name for consistency?

custodial_copy.tf
```diff
@@ -46,10 +46,10 @@ module "dr2_custodial_copy_db_builder_queue" {
 module "dr2_custodial_copy_queue" {
   source     = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
   queue_name = local.custodial_copy_name
```
techncl, line 48: I know you didn't add this in this PR but should we rename this slightly? The other one is called "dr2-custodial-copy-queue-creator" but since this is called "dr2-custodial-copy", it might get confused for the other in future, maybe?

custodial_copy.tf
MancunianSam: Oops, thanks.

common.tf
MancunianSam: It's also being used for the queue name which is why I didn't put lambda in there. If you look at the other lambdas that have queues like `local.copy_files_from_tdr_name` then they're the same.

custodial_copy.tf
MancunianSam: I see what you mean but the name isn't used for just the queue, it's used for the user and group as well so I'm not sure what we'd change it to.

techncl (approved):
👍

PR #188, opened 2024-10-04
Make 'Map over each Asset Id' a distributed map

techncl:


templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -91,15 +91,29 @@
...
       "Type": "Map",
+      "Label": "MapOverEachAssetId",
       "ItemsPath": "$.contentAssets",
       "ItemSelector": {
         "id.$": "$$.Map.Item.Value",
         "batchId.$": "$$.Execution.Input.batchId",
-        "executionName.$": "$$.Execution.Name"
+        "executionName.$": "$$.Execution.Name",
+        "bucket.$": "$.assets.bucket",
+        "key.$": "$.assets.key"
```
kurtismash, line 101: Why are the bucket and key needed here? I'm assuming this is the `assets.json` in the state bucket?

techncl: 😄 it's funny, initially, in the terraform, I just used the ExecutionName + `assets.json` for the key and add the bucket name but then I thought that the reason why you put "return the bucket and key" on the ticket was so that we didn't have to manually edit the terraform in case we changed the bucket/key in the future, so I changed it to this

kurtismash: You did the right thing! We are using the bucket and key from the state data in the ItemReader but steps within the map don't need to know about these as they update the DDB table or don't give a return value we care for.

kurtismash: Oh wait, what's `$.bucket` and `$.key` in `ItemReader`. I _think_ `ItemReader` happens on the Map state and `ItemSelector` defines the input to the child step functions.

techncl: Yes, `ItemSelector` filters the Input and then I'm using the `bucket` and `key`, from the input, in the `ItemReader`. When I was editing via the console, it added `Parameters` to the step definition so I kept it.

kurtismash: I don't think that's correct. From the [docs](https://docs.aws.amazon.com/step-functions/latest/dg/state-map-distributed.html#map-state-distributed-additional-fields),
- "ItemReader Specifies a dataset and its location. The Map state receives its input data from the specified dataset."
- "ItemSelector (Optional) Overrides the values of individual dataset items before they're passed on to each Map state iteration."

So the ItemReader here would fail to find the object as `$.bucket` and `$.key` don't exist.

techncl: Ah, I see.

kurtismash (approved):

PR #190, opened 2024-10-04
Make 'Map over each Folder Id' a distributed map

techncl:


templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -201,23 +201,42 @@
...
+        "assetsKey.$": "$.assetsKey",
         "contentAssets.$": "$.contentAssets",
         "executionId.$": "$$.Execution.Name",
+        "foldersBucket.$": "$.foldersBucket",
+        "foldersKey.$": "$.foldersKey",
         "workflowContextName": "Ingest OPEX (Incremental)",
         "stagingPrefix.$": "States.Format('opex/{}', $$.Execution.Name)",
         "stagingBucket": "${ingest_staging_cache_bucket_name}"
       }
     },
```
kurtismash, line 226: We should be able to remove the `Convert array to object` and `Merge array` states as these were used to join the `archiveFolders` and `contentFolders` arrays, which we've now already done within the Mapper Lambda when generating `folders.json`.

templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -201,23 +201,42 @@
...
+      "Label": "MapOverEachFolderId",
+      "ItemReader": {
+        "Resource": "arn:aws:states:::s3:getObject",
+        "ReaderConfig": {
+          "InputType": "JSON"
+        },
+        "Parameters": {
+          "Bucket": "$.foldersBucket",
+          "Key.$": "$.foldersKey"
+        }
```
kurtismash, line 238: See other comment, this should make these `$.folders.bucket` and `$.folders.key` respectively. Nice and simple :)

templates/sfn/ingest_sfn_definition.json.tpl
techncl: Oh yeah, Cool

templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -190,34 +190,22 @@
...
+        "Resource": "arn:aws:states:::s3:getObject",
+        "ReaderConfig": {
+          "InputType": "JSON"
+        },
+        "Parameters": {
+          "Bucket.$": "$.folders.bucket",
+          "Key.$": "$.folders.key"
+        }
+      },
       "ItemsPath": "$.allFolders",
```
kurtismash, line 209: We need to remove `ItemsPath` as we're now getting our data from S3 instead of the state data. The same will need doing on the asset map.

templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -392,7 +384,10 @@
       "Type": "Task",
       "Resource": "arn:aws:states:::lambda:invoke",
       "Parameters": {
-        "Payload.$": "$",
+        "Payload": {
+          "executionId.$": "$$.Execution.Name",
+          "contentAssets.$": "$.contentAssets"
```
kurtismash, line 389: Oh, I forgot we did this. We'll need to make changes here as it'll be a bottleneck for the number of assets in a batch. I'll put some tickets in.

kurtismash (approved):

PR #191, opened 2024-10-07
DR2-1910 Add ingest failure notifications lambda.

MancunianSam:
This needs
https://github.com/nationalarchives/da-terraform-modules/pull/71 before
the tests will pass but it's ready for review.


techncl (approved):
👍

PR #192, opened 2024-10-08
Reconcile on workflow failure.

MancunianSam:
If the Preservica workflow fails, some assets may have been ingested and
we want these to progress and send their complete messages which they
won't if the step function fails.

The reconciler will return errors for the ones that haven't ingested.

I've reused the wait 5 minutes step just to tidy things up a bit. Most
ingests happen well before that and if not, another 5 minutes isn't
going to matter much.

I've also shortened the name of the state because the really wide boxes
make it difficult to see what's going on.

### DIFF
--- a/templates/sfn/ingest_sfn_definition.json.tpl
+++ b/templates/sfn/ingest_sfn_definition.json.tpl
@@ -425,28 +425,26 @@
         "duplicatedAssets.$": "$.Payload.duplicatedAssets"
       },
       "ResultPath": "$.WorkflowResult",
-      "Next": "Check workflow status and get Succeeded, Failed and Duplicated asset ids"
+      "Next": "Check workflow status"
     },
-    "Check workflow status and get Succeeded, Failed and Duplicated asset ids": {
+    "Check workflow status": {
       "Type": "Choice",
       "Choices": [
         {
-          "Variable": "$.WorkflowResult.status",
-          "StringEquals": "Failed",
-          "Next": "Job Failed"
-        },
-        {
-          "Variable": "$.WorkflowResult.status",
-          "StringEquals": "Succeeded",
+          "Or": [
+            {
+              "Variable": "$.WorkflowResult.status",
+              "StringEquals": "Failed"
+            },
+            {
+              "Variable": "$.WorkflowResult.status",
+              "StringEquals": "Succeeded"
+            }
+          ],
           "Next": "Map over each assetId and reconcile"
         }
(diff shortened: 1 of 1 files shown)


templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -425,28 +425,26 @@
         "duplicatedAssets.$": "$.Payload.duplicatedAssets"
       },
       "ResultPath": "$.WorkflowResult",
-      "Next": "Check workflow status and get Succeeded, Failed and Duplicated asset ids"
+      "Next": "Check workflow status"
     },
-    "Check workflow status and get Succeeded, Failed and Duplicated asset ids": {
+    "Check workflow status": {
```
techncl, line 430: I just realised that this step and the previous one weren't originally given the most accurate names
   1. Maybe the previous step could be (in order to emphasise that we are returning ids)  "Get workflow status & asset id per status" or "Get workflow & asset id statuses"
   2. and then this step could be named "Check for Failed or Succeeded workflow status" or "Check workflow status for Failed or Succeeded" (even though you could get that from reading the step definition, I was thinking from a non-technical person's perspective)?

templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -425,28 +425,26 @@
...
       ],
-      "Default": "Wait 1 minute"
-    },
-    "Wait 1 minute": {
-      "Type": "Wait",
-      "Next": "Get workflow status",
-      "Seconds": 60
+      "Default": "Wait 5 minutes before getting status"
     },
     "Job Failed": {
```
techncl, line 449: Should this be removed?

techncl (commented):
Looks good! Just two comments

templates/sfn/ingest_sfn_definition.json.tpl
MancunianSam: I don't think we're going to get many non-technical people reading our step functions and even if there were, "Check workflow status" conveys broadly what it's doing. I would prefer to keep the names short and as you say, the detail can be found in the step function details.

templates/sfn/ingest_sfn_definition.json.tpl
MancunianSam: No, I tried, the batch workflow status check is using it.

techncl (approved):
👍

PR #194, opened 2024-10-08
Make 'Map over each assetId and reconcile' a distributed map

techncl:


templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -440,22 +445,29 @@
...
       "Type": "Map",
-      "ItemsPath": "$.contentAssets",
+      "Label": "MapOverEachAssetIdAndReconcile",
+      "ItemReader": {
+        "Resource": "arn:aws:states:::s3:getObject",
+        "ReaderConfig": {
+          "InputType": "JSON"
+        },
+        "Parameters": {
+          "Bucket": "$.assets.bucket",
```
MancunianSam, line 457: This needs to be "Bucket.$" otherwise it's trying to read from a bucket called "$.assets.bucket"

MancunianSam (approved):

PR #204, opened 2024-10-28
Step function updates

MancunianSam:
The distributed map which creates the asset opex needs to be batched otherwise we run out of lambda invocations.

The reconciler needs to stop outputing every failed child to Slack as this won’t work if there’s more than a few of them.

The reconciler also needs to send its output to S3 as it’s too large at the moment.

There are a few permissions updates needed and some of the SFN policy is out of date.

### DIFF
--- a/templates/iam_policy/ingest_step_function_policy.json.tpl
+++ b/templates/iam_policy/ingest_step_function_policy.json.tpl
@@ -10,8 +10,7 @@
         "sts:AssumeRole",
         "events:PutEvents",
         "dynamodb:Query",
-        "dynamodb:DeleteItem",
-        "sns:Publish"
+        "dynamodb:DeleteItem"
       ],
       "Resource": [
         "arn:aws:lambda:eu-west-2:${account_id}:function:${ingest_validate_generic_ingest_inputs_lambda_name}",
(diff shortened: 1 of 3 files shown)


templates/iam_policy/ingest_step_function_policy.json.tpl
```diff
@@ -27,9 +27,8 @@
         "arn:aws:dynamodb:eu-west-2:${account_id}:table/${ingest_lock_table_name}",
         "arn:aws:dynamodb:eu-west-2:${account_id}:table/${ingest_lock_table_name}/index/${ingest_lock_table_group_id_gsi_name}",
         "arn:aws:states:eu-west-2:${account_id}:stateMachine:${ingest_sfn_name}",
-        "arn:aws:states:eu-west-2:${account_id}:execution:${ingest_sfn_name}/StagingCacheS3ObjectKeys:*",
-        "arn:aws:states:eu-west-2:${account_id}:execution:${ingest_sfn_name}:*",
-        "arn:aws:sns:eu-west-2:${account_id}:${notifications_topic_name}",
+        "arn:aws:states:eu-west-2:${account_id}:execution:intg-dr2-ingest/MapOverEachFolderId:*",
```
techncl, line 30: 1. Should `intg-dr2-ingest` not be `${ingest_sfn_name}`?
2. Just out of curiosity, what is this resource? The Distributed Map?

templates/sfn/ingest_sfn_definition.json.tpl
```diff
@@ -553,7 +539,7 @@
               "Entries": [
                 {
                   "Detail": {
-                    "slackMessage.$": "$.reason"
+                    "slackMessage.$": ":alert-noflash-slow: Reconcile failed for asset $.assetId"
```
techncl, line 542: 1. Maybe "Reconciliation" rather than "Reconcile"?
2. Is it possible/Should we add the bucket/Object path to this message so that the output could be found easier?

techncl (commented):
Looks good! Just a few minor comments. Also, can the `sns:Publish` action be removed?

templates/iam_policy/ingest_step_function_policy.json.tpl
MancunianSam: 1. Yep, thanks. Copy and paste fail.
2. It is yeah, if there's a failure inside the distributed map, it lets us do a redrive from that failure.

templates/sfn/ingest_sfn_definition.json.tpl
MancunianSam: 1. Ok
2. It's a bit fiddly. The path in S3 of the results is `reconcilerOutput/<map-run-uuid>` but it's not trivial to get that ID, you have to set it yourself and then pass it through with the standard input which seems like overkill. I'll put a line in the slack message to say "See the state output for the result key." and we'll see how we get on with that.

techncl (approved):
👍

PR #205, opened 2024-11-05
Increase max session duration.

MancunianSam:
I suspect that the tests weren't taking over an hour, they were just
broken but they might if we start chucking bigger ingests through so I'm
going to add this in anyway.


techncl (commented):
Looks good! We've used 7200 seconds twice in this file (on line 47 also); do we want to save this to a variable so that they are in sync?

MancunianSam:
Hmm, it feels a bit like overkill. We could have a variable called `two_hours` but then if we have two things that need three hours do we add `three_hours` It's ok for the lambdas as that's a default of sorts but I don't think a variable would add much here.

techncl (approved):
👍

PR #206, opened 2024-11-07
Remove preservica config lambda.

MancunianSam:
The code is being removed https://github.com/nationalarchives/dr2-ingest/pull/339

We need to remove the lambda after that has deployed.


techncl (approved):
👍

PR #207, opened 2024-11-07
Fix Step Function errors

techncl:


templates/sfn/ingest_run_workflow_sfn_definition.json.tpl
```diff
@@ -122,10 +122,14 @@
...
           ],
-          "Next": "Map over each assetId and reconcile"
+          "Next": "End Workflow and return WorkflowResult"
         }
       ],
       "Default": "Wait 5 minutes before getting status"
+    },
+    "End Workflow and return WorkflowResult": {
+      "Type": "Pass",
+      "End": true
```
kurtismash, line 132: I think we can save a state transition by making this a `Succeed` state instead of a `Pass`.

https://docs.aws.amazon.com/step-functions/latest/dg/state-succeed.html

kurtismash (dismissed):

templates/sfn/ingest_run_workflow_sfn_definition.json.tpl
techncl: But can a `Succeed` state pass on its Output?

kurtismash: Good question! It ends the execution so I'd assume it does, in the same way that `"End": true` does in the background.

kurtismash (approved):

PR #215, opened 2024-11-28
DR2-2011 Replace SNS with EventBridge

MancunianSam:
We're not sending an SNS message any more, we're sending a message to
EventBridge.


techncl (commented):
👍

techncl (approved):

PR #216, opened 2024-11-28
DR2-1999 Add missing lambdas to dashboard

MancunianSam:
I can find 23 lambdas in here so I've added them all in.


techncl (commented):
Looks good. All of these lambdas will be used for the "Ingest dashboard", should we maybe have multiple lamba dashboards: Ingest, PreIngest, Preservica? Maybe it would make it easier to find things also?

MancunianSam:
We might do that in the future. I don't want to spend time doing it now though until we know who is using these dashboards (if anyone) and whether that would be helpful. 
At the moment, we tend to get errors in one or two places because those errors block anything downstream happening so I think it's ok to have them all lumped together for now. 

In short, maybe but not yet.

techncl (approved):
👍

PR #217, opened 2024-11-29
DR2-2008 Update provider

MancunianSam:
I've removed the assume roles section as we'll be assuming the role in
the GitHub action.

I've also removed the account number variable as we don't need it and
replaced everything inside with the data.aws_caller_identity block.


techncl (approved):
👍

PR #219, opened 2025-01-13
DR2-1948 Delete staging cache bucket and provider.

MancunianSam:
The changes to remove datasync are in prod so we don't need this any
more.

There'll be a manual step to empty the bucket before this can be
deployed.


techncl (approved):
👍

PR #220, opened 2025-01-13
DR2-2034 Rename lambda policies

MancunianSam:
We have a wildcard on the terraform policy for iam:* with a filter of
<env>* on role and policy names.
These names didn't match this pattern this will make them all match.


techncl (approved):
👍

PR #222, opened 2025-01-30
Remove daily alarm and increase thresholds.

MancunianSam:
The daily alarm is not that helpful so that's being deleted.

The other custodial copy ones have had their threshold raised to 50
because it's fairly common that we get more than that for a single
ingest.

This has already been deployed.


techncl (approved):
👍

PR #224, opened 2025-02-05
DR2-2086 Configure TDR Aggregator for larger message bursts

kurtismash:
Configures the TDR Aggregation workflow to handle larger bursts of messages. It's unknown how many messages the Lambda can handle in a single invocation before timeout so I've reduced this for now - it's better to have multiple invocations build a group than an invocation timeout as the messages will keep looping in Lambda/SQS until the ingest for that asset is unlocked.


kurtismash:
Tested manually on intg, the Lambda can process >6k messages within 30 seconds - although the config can pass <=10k messages into an invocation, this won't happen due to the mandatory SQS metadata that comes with each message.

Not tested if the Package Builder Lambda can handle a group of 10k records.

MancunianSam (approved):

PR #233, opened 2025-04-08
DR2-2190 Use correct input param.

MancunianSam:
The lock table contains the groupId, not the batchId, so this query
always returns zero results.


techncl (approved):
👍

PR #234, opened 2025-04-08
DR2-2116 Remove datasync resources.

MancunianSam:
* Deleted the log group
* Removed datsync from the Cloudwatch resource policy.


techncl (approved):
👍

PR #236, opened 2025-04-16
Name change to avoid clash.

MancunianSam:
This was clashing with
`"${local.environment}-dr2-eventbridge-alarm-state-change-${lower(each.value)}"`
which comes out as the same thing when each.value is ALARM.


techncl (approved):
👍

PR #239, opened 2025-04-17
DR2-2139 Configure Ingest Flow Control

kurtismash:
We have 2 environments and a single Job Queue server with 16 threads. We've been advised to use 10 threads in total for our activities, so limiting to 5 concurrent ingests per environment.


techncl (approved):
👍

PR #240, opened 2025-04-28
DR2-2139 Remove dedicated channel to test probability weighting

kurtismash:


techncl (commented):
Looks good but shouldn't court docs have a higher number than TDR since it's more important they they go through promptly?

techncl (approved):
👍

PR #241, opened 2025-04-30
Dr2 2205 reporting infrastructure terraform

sparkhi:
Added the terraform infrastructure 
- S3 bucket, lambda, cloudwatch eventbridge, cloudfront

### DIFF
--- a/common.tf
+++ b/common.tf
@@ -3,7 +3,7 @@ locals {
   ingest_raw_cache_bucket_name                         = "${local.environment}-dr2-ingest-raw-cache"
   sample_files_bucket_name                             = "${local.environment}-dr2-sample-files"
   ingest_state_bucket_name                             = "${local.environment}-dr2-ingest-state"
-  reporting_bucket_name                                = "${local.environment}-dr2-reporting"
+  ingest_reporting_bucket_name                         = "${local.environment}-dr2-ingest-reporting"
   ingest_step_function_name                            = "${local.environment}-dr2-ingest"
   ingest_run_workflow_step_function_name               = "${local.environment}-dr2-ingest-run-workflow"
   additional_user_roles                                = local.environment != "prod" ? [data.aws_ssm_parameter.dev_admin_role.value] : []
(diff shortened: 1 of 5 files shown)


reporting_dashboard.tf
```diff
@@ -0,0 +1,73 @@
+locals {
+  reporting_helper_lambda_name = "${local.environment}-dr2-reporting-helper"
```
techncl, line 2: This name seems a bit vague; I called mine "ingest timings". Maybe we can call it something like "dr2-ingest-timings-report-lambda"?

reporting_dashboard.tf
```diff
@@ -0,0 +1,73 @@
...
+  policies = {
+    "${local.reporting_helper_lambda_name}-policy" = templatefile("./templates/iam_policy/reporting_helper_policy.json.tpl", {
+      account_id  = data.aws_caller_identity.current.account_id
+      lambda_name = local.reporting_helper_lambda_name
+      bucket_name = local.reporting_bucket_name
+    })
+  }
+  memory_size = local.python_lambda_memory_size
+  runtime     = local.python_runtime
+
```
techncl, line 28: Could you add the environment variables here? I've added them to [my PR](https://github.com/nationalarchives/dr2-ingest/pull/483/files#diff-930007e96c8ccb8cee7ea72cab45ab3f1a5413e0328eba5b30dedb8c0d378bbbR43)

templates/iam_policy/reporting_helper_policy.json.tpl
```diff
@@ -0,0 +1,28 @@
+{
+  "Statement": [
+    {
+      "Action": [
+        "s3:GetObject",
```
techncl, line 5: I'm not sure it needs this permission

common.tf
```diff
@@ -3,6 +3,7 @@ locals {
   ingest_raw_cache_bucket_name                         = "${local.environment}-dr2-ingest-raw-cache"
   sample_files_bucket_name                             = "${local.environment}-dr2-sample-files"
   ingest_state_bucket_name                             = "${local.environment}-dr2-ingest-state"
+  reporting_bucket_name                                = "${local.environment}-dr2-reporting"
```
techncl, line 6: 1. Should this variable name start with "ingest"
2. Should "ingest" appear in the name (after "dr2")?

techncl (commented):
Looks good; just a few comments. It's probably best that someone else looks at this as well since I'm not too knowledgable on CloudFront.

reporting_dashboard.tf
sparkhi: I thought this would be multiple metrics that we want to surface to dashboard. Ingest timing is just the first one. Maybe I can simply change it to "dr2-ingest-reporting"

templates/iam_policy/ingest_reporting_policy.json.tpl
```diff
@@ -2,8 +2,7 @@
   "Statement": [
     {
       "Action": [
-        "s3:GetObject",
-        "s3:PutObject"
+        "s3:GetObject"
```
techncl, line 5: You removed `s3:PutObject` instead of  `s3:GetObject`

MancunianSam (approved):
Looks good thanks. This has already been run I think so it should work.

MancunianSam (approved):

PR #245, opened 2025-05-23
Use assetId not ioRef

MancunianSam:
We can't use ioRef for the PK because the output messages need to have
our reference, not Preservicas.

The confirmer job will need to contact Preservica to get their reference
from ours


techncl (approved):
👍

PR #246, opened 2025-05-27
Add ioRef as well.

MancunianSam:
This way, the change handler can pass this to the CC confirmer which
will keep things much simpler.


techncl (approved):
👍

PR #247, opened 2025-05-30
DR2-2258 Remove PRESERVICA_API_URL

MancunianSam:
Once https://github.com/nationalarchives/dr2-ingest/pull/513 is merged
and deployed, this can be run.

We're now getting the API url from the secret so there's no need to have
it here


techncl (approved):
👍

PR #248, opened 2025-06-05
Dr2 2254 update post ingest terraform

techncl:


post_ingest/post_ingest.tf
```diff
@@ -70,6 +73,12 @@ module "dr2_state_change_lambda" {
   dynamo_stream_config = {
     stream_arn = module.post_ingest_state_table.stream_arn
   }
+  plaintext_env_vars = {
+    POST_INGEST_STATE_DDB_TABLE                = local.post_ingest_state_table_name
+    POST_INGEST_DDB_TABLE_BATCHPARENT_GSI_NAME = local.post_ingest_gsi_name
```
kurtismash, line 78: Can we use `POSTINGEST` to match [`PREINGEST`](https://github.com/nationalarchives/dr2-terraform-environments/blob/c993791044208e08dee6c6de0cb2367a9ae4a368/tdr_preingest.tf#L61C5-L61C22) please.

post_ingest/post_ingest.tf
```diff
@@ -70,6 +73,12 @@ module "dr2_state_change_lambda" {
   dynamo_stream_config = {
     stream_arn = module.post_ingest_state_table.stream_arn
   }
+  plaintext_env_vars = {
+    POST_INGEST_STATE_DDB_TABLE                = local.post_ingest_state_table_name
```
kurtismash, line 77: Consider removing "STATE", we're referring to this table as both `POST_INGEST_STATE_DDB_TABLE` and `POST_INGEST_DDB_TABLE_BATCHPARENT_GSI_NAME`.

post_ingest/post_ingest.tf
```diff
@@ -6,6 +6,9 @@ locals {
   resender_lambda_name                = "${var.environment}-dr2-postingest-message-resender"
   java_runtime                        = "java21"
   java_lambda_memory_size             = 512
+  post_ingest_queue_config = [
+    { "queueAlias" : "CC", "queueOrder" : 1, "queueUrl" : module.dr2_custodial_copy_confirmer_queue.sqs_queue_url }
```
kurtismash, line 10: Do have handling within the Lambda for multiple queues with the same alias?

post_ingest/templates/policies/state_change_lambda_policy.json.tpl
```diff
@@ -22,13 +22,14 @@
     {
       "Action": [
         "dynamodb:BatchWriteItem",
-        "dynamodb:PutItem"
+        "dynamodb:PutItem",
```
kurtismash, line 25: I think we want to be using [`UpdateItem`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html) instead to avoid losing an attribute in a race condition.

postingest/outputs.tf
```diff
@@ -0,0 +1,11 @@
...
+  value = local.postingest_state_table_name
+}
+
+output "postingest_table_arn" {
+  value = module.postingest_state_table.table_arn
+}
+
+output "postingest_confirmer_queue_arn" {
+  value = module.dr2_custodial_copy_confirmer_queue.sqs_arn
+}
```
MancunianSam, line 11: You'll need to output the lambda arn here as well and then add it to the list in `deploy_lambda_policy.tf` The deploy job is failing at the moment.

MancunianSam: You've got it in the output, you just need to add the output to the deploy role.
So add `module.postingest.postingest_state_change_lambda_arn` to the bottom of the list here 
https://github.com/nationalarchives/dr2-terraform-environments/blob/main/deploy_roles.tf#L51

techncl: Oops, I missed the 2nd half of the sentence.

MancunianSam (approved):

PR #252, opened 2025-07-02
Dr2 2252 tidy up after postingest

sparkhi:
**NOTE**: This removed the SQS queue and Lambda for CC notifications. This should not to be merged before Postingest other work is done.


MancunianSam (approved):
:+1:

MancunianSam:
You can also remove the entire file here https://github.com/nationalarchives/dr2-terraform-environments/blob/main/ingest_files_change_handler.tf as we're not using that any more.

MancunianSam (approved):

PR #255, opened 2025-07-09
Fixes for resender

MancunianSam:
* Add the environment variables to the lambda
* Add missing permissions to the lambda policy


techncl (approved):
👍

PR #259, opened 2025-07-22
DR2-2317 Make preingest a module.

MancunianSam:
This moves the following into a module.
* The package builder, aggregator and copy files lambdas
* The associated queues
* The preingest step function

This way, we can add a new preingest source just by adding another
module call into preingest.tf at the root.


techncl (approved):
👍

PR #261, opened 2025-07-28
Add source system env var

MancunianSam:
This will allow us to have two metadata schemas, one for DRI and one for
TDR.


techncl (approved):
👍

PR #271, opened 2025-09-16
DR2-2404 Alarm if oldest confirmer message more than a week.

MancunianSam:


techncl (commented):
Looks good! Just some minor comments.

PR #279, opened 2025-10-06
DR2-2492 Add Cloudflare outbound security group.

MancunianSam:
This needs
https://github.com/nationalarchives/da-terraform-modules/pull/128 before
the tests will pass.

### DIFF
--- a/common.tf
+++ b/common.tf
@@ -161,18 +161,22 @@ module "outbound_https_access_only" {
   description = "A security group to allow outbound access only"
   name        = "${local.environment}-outbound-https"
   vpc_id      = module.vpc.vpc_id
-  egress_cidr_rules = [{
-    port        = 443
-    description = "Outbound https access",
-    cidr_blocks = ["0.0.0.0/0"],
-    protocol    = "tcp"
-  }]
-  egress_security_group_rules = [{
-    port              = 443
-    description       = "Outbound https to discovery VPC endpoint"
-    security_group_id = module.discovery_inbound_https.security_group_id
-    protocol          = "tcp"
-  }]
+  rules = {
+    egress = [
+      {
+        port              = 443
+        description       = "Outbound https to discovery VPC endpoint"
+        security_group_id = module.discovery_inbound_https.security_group_id
+        protocol          = "tcp"
+      },
+      {
+        port        = 443
+        description = "Outbound https to all IPs"
+        cidr_ip_v4  = "0.0.0.0/0"
+        protocol    = "tcp"
+      },
+    ]
+  }
 }
 
 resource "aws_ec2_managed_prefix_list" "cloudflare_prefix_list" {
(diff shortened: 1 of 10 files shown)


common.tf
```diff
@@ -74,6 +74,7 @@ locals {
   ]
   retry_statement            = jsonencode([{ ErrorEquals = ["States.ALL"], IntervalSeconds = 2, MaxAttempts = 6, BackoffRate = 2, JitterStrategy = "FULL" }])
   messages_visible_threshold = 1000000
+  cloudflare_ip_ranges       = toset(["173.245.48.0/20", "103.21.244.0/22", "103.22.200.0/22", "103.31.4.0/22", "141.101.64.0/18", "108.162.192.0/18", "190.93.240.0/20", "188.114.96.0/20", "197.234.240.0/22", "198.41.128.0/17", "162.158.0.0/15", "104.16.0.0/13", "104.24.0.0/14", "172.64.0.0/13", "131.0.72.0/22"])
```
kurtismash: Worth adding a comment here to https://www.cloudflare.com/en-gb/ips as it'll be something to check during debugging.

common.tf
```diff
@@ -174,6 +175,32 @@ module "outbound_https_access_only" {
   }]
 }
 
+resource "aws_ec2_managed_prefix_list" "cloudflare_prefix_list" {
+  address_family = "IPv4"
+  max_entries    = 15
```
kurtismash: `length(local.cloudflare_ip_ranges)`?

Does this require replacement when changed?

common.tf
```diff
@@ -174,6 +175,32 @@ module "outbound_https_access_only" {
...
+    content {
+      cidr = entry.value
+    }
+  }
+}
+
+module "outbound_cloudflare_https_access" {
+  source      = "git::https://github.com/nationalarchives/da-terraform-modules//security_group"
+  common_tags = {}
+  description = "A security group to allow outbound access only"
```
kurtismash: Add Cloudflare to description.

kurtismash (commented):
Looks good, suggested a few minor changes to help future explorers, and one that may catch us out down the line.

custodial_copy_queue_creator.tf
```diff
@@ -37,7 +37,7 @@ module "dr2_custodial_copy_queue_creator_lambda" {
   }]
   vpc_config = {
     subnet_ids         = module.vpc.private_subnets
-    security_group_ids = [module.outbound_https_access_only.security_group_id]
+    security_group_ids = [module.outbound_https_access_only.security_group_id, module.outbound_cloudflare_https_access.security_group_id]
```
techncl: Should we maybe add this to the common.tf file?

common.tf
MancunianSam: It would have done, I've changed it to use length.

custodial_copy_queue_creator.tf
MancunianSam: Yeah good point, I've made a local variable.

common.tf
kurtismash: Would it be better to hardcode it with some leeway in that case? If we need to add a new range will this cause downtime otherwise?

MancunianSam: Yeah OK, I've added 5 to the length, that feels like enough.

kurtismash (approved):

